PR #172, opened 2026-05-24
fix(env): unify .env secrecy model + provenance gate for tool-managed files

Kostiantyn-Rozbey:
## Summary

- Introduces `src/utils/env-classifier.ts` as a single source of truth for whether a `.env`-shaped file is `template`, `managed`, or `user-secret`. The shell guard (`hooks/trace-mcp-guard.sh`), the `get_env_vars` MCP tool, and the indexer were each carrying their own version of this rule; they disagreed on tool-emitted files such as `~/.trace-mcp/launcher.env` and produced false positives in the guard's coach advisory.
- Adds a provenance gate: a `.env` file whose first non-blank line is `# Managed by <name>` with `<name>` in a **closed allowlist** (currently just `trace-mcp`) is treated as `managed` and exempt from the advisory. The marker is a trust signal about the emitter, not a content guarantee — a forged `# Managed by attacker` stays classified as `user-secret`.
- Exposes the classifier to the shell guard via a tiny `trace-mcp classify-env <path>` subcommand that prints one line of JSON. The guard falls back to its existing filename regex if the CLI is missing, returns non-JSON, or the path is a pattern rather than a real file — so the hook never becomes *less* strict on classifier failure.
- Adds opt-in `redacted: true` mode to `get_env_vars` (requires `file`) that pipes the named file through the existing `redactEnvFile()` util, returning a line-by-line view with values replaced by type hints. Useful when key ordering and comments matter. Default behaviour is unchanged.

## Motivation

While diagnosing an unrelated bug I read `~/.trace-mcp/launcher.env` and got hit with the `[trace-mcp coach] Use get_env_vars for .env files…` advisory — even though that file is generated by trace-mcp's own postinstall, starts with the comment \`# Managed by trace-mcp postinstall — do not edit by hand.\`, and holds only paths and a version string. The hook decides purely on filename, so it can't distinguish a user's \`.env\` containing AWS keys from trace-mcp's own service file. False positives like this train users to ignore the advisory, which is the worst possible outcome for a security hint. See the architectural rationale in the commit message.

## Why this approach

- **Lights up an unused primitive.** \`redactEnvFile()\` from \`src/utils/env-parser.ts\` already produces \`KEY=<string:dsn>\` views; it was only called by the source-reader. Surfacing it through \`get_env_vars { redacted: true }\` gives callers a useful middle ground between \"names only\" and \"raw values\" without writing any new redaction code.
- **Reduces drift across three secrecy models.** \`is_sensitive_env_file\` (shell), the indexer's \`isSensitiveFile\`, and \`get_env_vars\` no longer need to stay synchronised by hand — the classifier is the single rule.
- **Doesn't relax security.** Allowlisted emitters are exact-string matches; the marker only excuses files emitted by a known tool. Pure user \`.env\` files are still caught.
- **Localised change.** No new package, no harness-side changes; everything lives in this repo.

## Test plan

- [x] \`pnpm exec vitest run tests/utils/env-classifier.test.ts\` — 14 new classifier tests, all pass.
- [x] \`pnpm exec vitest run tests/utils/env-parser.test.ts tests/tools/behavioural/get-env-vars.behavioural.test.ts tests/indexer/env-indexer.test.ts\` — 35 tests in modified areas, all pass.
- [x] \`pnpm exec tsc --noEmit\` — error count identical to master (474 lines); this PR introduces no new typecheck errors.
- [x] \`bash -n hooks/trace-mcp-guard.sh\` — syntax valid.
- [x] End-to-end CLI smoke:
  - \`trace-mcp classify-env\` on a fixture with the \`# Managed by trace-mcp\` header → \`{\"tier\":\"managed\"}\`
  - on a fixture with \`AWS_SECRET_KEY=...\` → \`{\"tier\":\"user-secret\"}\`
  - on \`.env.example\` → \`{\"tier\":\"template\"}\`
  - on a non-existent path → \`{\"tier\":\"user-secret\"}\` (filename-only fallback)
- [x] Shell JSON extraction (\`sed -n 's/.*\"tier\"...'\`) returns the correct tier for all four CLI outputs above.

## Open questions for reviewers

- Should \`redacted: true\` become the **default** for \`get_env_vars\` (with \`redacted: false\` for the current names-only view)? This PR keeps the current default to preserve the public contract; happy to flip if you'd rather.
- The \`MANAGED_EMITTER_ALLOWLIST\` is currently \`[\"trace-mcp\"]\`. Extending it is a security decision per entry — the emitter must guarantee its file holds no real secrets. I deliberately did not add other ecosystem tools (Docker Compose, Vercel CLI, etc.) so this PR stays scope-tight.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nikolai-vysotskyi (approved):
Code review LGTM after my follow-up fixes:

- Original env-classifier work is well-designed: closed allowlist, exact-string emitter match, first-non-blank-line scan prevents header smuggling, graceful filename-only fallback when head bytes unavailable.
- Verified locally: build + 49 tests pass, typecheck identical to master (474 lines), CLI smoke (managed/secret/template/forged/nonexistent) all classify correctly, shell-syntax check passes.
- Pushed three follow-ups: (1) biome format fixes on cli.ts and env-classifier.ts, (2) path-injection hardening on `get_env_vars { redacted: true }` after CodeQL flagged user-controlled `file` flowing into fs.readFileSync — final form uses startsWith(projectRoot+sep) containment + indexed-allowlist gate; CodeQL is satisfied.
- All 7 CI workflows green (CI, CodeQL Analyze, Semgrep, Eval, Semgrep OSS, GHAS CodeQL rollup).

PR #173, opened 2026-05-24
fix(app): stop in-app updater from looping on npm-only outcomes

Kostiantyn-Rozbey:
## Summary

- The Update button looped: \`apply-update\` ran \`npm install -g trace-mcp@latest\`, then claimed \`pending: true\` whenever the on-disk npm package was newer than the running Electron process. The UI showed "Restart to install", the user restarted, the \`.app\` bundle had **not** actually been swapped (npm install cannot touch \`/Applications/trace-mcp.app\`), and \`check-for-update\` saw the same mismatch again. Plashka returned indefinitely.
- This PR splits the apply-update result into three honest outcomes via a new pure module ([packages/app/src/main/update-state.ts](https://github.com/Kostiantyn-Rozbey/trace-mcp/blob/fix/update-loop-pending-truth/packages/app/src/main/update-state.ts)):
  - **bundle-pending** — a verified zip is staged; restart will swap the bundle. Drives \`pending: true\`.
  - **npm-only** — CLI moved but the Electron bundle is stuck. \`pending: false\`. A sticky marker is written to \`~/.trace-mcp/app-update-state.json\` so \`check-for-update\` returns \`available: false, stuck: true\` until either the registry advances past the stuck target or the user manually reinstalls the .app.
  - **already-current** — nothing to do.
- \`check-pending-update\` drops its legacy "npm-install path" branch — that branch returned \`pending: true\` based on the in-memory \`pendingRestartVersion\` variable, which was unsound (relaunch cannot move the bundle) and was the proximate cause of the spurious "Restart to install" prompt.
- Renderer refreshes its banner state immediately after a successful apply-update, so the user sees the loop break on the same click instead of waiting for the 10-minute poll.

## Reproduction

User-reported symptom: clicking Update → \"Update to restart\" → restart → plashka for the same version comes back. Visible in [~/.trace-mcp/update.log](#) as repeated \`{event:'apply-update:ok', pending:true, installedVersion:'X', runningVersion:'Y'}\` where Y stays the same across attempts.

## Why this approach

- **Tells the truth about \"pending\".** The previous OR (\`hasPendingUpdate() || pendingRestartVersion !== undefined\`) conflated \"bundle swap queued\" with \"npm package newer on disk\". Only the first is fixable by restart.
- **Doesn't add a permanent suppression.** The sticky marker is keyed on \`(bundle, target)\`. A real new release (registry advances) immediately re-enables the banner; a manual reinstall (bundle moves) immediately re-enables the banner. Both auto-clear paths are tested.
- **Minimal renderer change.** One extra \`runCheck()\` call when \`outcome === 'npm-only'\` collapses the "Update available" card to "Up to date" on the same click — no new UI components.

## Test plan

- [x] \`pnpm exec vitest run tests/app-update-state.test.ts\` — 14 new pure-logic tests, all pass:
  - \`computeUpdateOutcome\`: bundle-pending overrides versions; npm-only when installed > running && no zip; already-current for equal/missing/regressed installedVersion.
  - \`isStuckOnVersion\`: no-marker false; exact match true; registry-advanced false; bundle-moved false; latest-regresses to target still true.
  - \`read/writeAppUpdateState\`: missing-file → {}; malformed-JSON → {}; round-trip; creates nested parent directories.
- [x] \`pnpm exec vitest run tests/updater.test.ts tests/updater-rollback.test.ts tests/app-update-state.test.ts\` — 31 tests pass, no regressions.
- [x] Electron main typecheck (\`tsc -p tsconfig.main.json --noEmit\`) — clean.
- [x] Full app typecheck (\`tsc --noEmit\` in packages/app) — clean.

## Files

- New: \`packages/app/src/main/update-state.ts\`, \`tests/app-update-state.test.ts\`
- Modified: \`packages/app/src/main/index.ts\`, \`packages/app/src/main/preload.ts\`, \`packages/app/src/renderer/App.tsx\`, \`packages/app/src/renderer/electron.d.ts\`

## Related

Companion PR [#172](https://github.com/nikolai-vysotskyi/trace-mcp/pull/172) addresses a separate issue (unified \`.env\` secrecy model + provenance gate). The two PRs are independent and can land in either order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nikolai-vysotskyi:
recheck

PR #243, opened 2026-07-14
feat: add gcf output format

blackwell-systems:
Adds `gcf` to the `output_format` enum alongside `json`, `markdown`, and `toon`. One branch in the shared `encodeResponse` helper, mirroring the existing TOON path (same fallback-to-JSON safety). Opt-in; `json` stays the default and TOON is untouched.

## Correctness

For a code-intelligence server this is the property that matters most: an agent acting on symbol/edge data cannot tolerate silent corruption. **GCF round-trips losslessly by design**, continuously fuzzed and verified (40B+ published round-trips).

By contrast, a 100k-case fuzz of the reference `@toon-format/toon` shows a **7.47% round-trip failure rate — 1.73% of it silent data corruption** (wrong value returned, no error thrown), the rest decode crashes. Reproducible: [`eval/toon-fuzz.mjs`](https://github.com/blackwell-systems/gcf/blob/main/eval/toon-fuzz.mjs) · [10M-case log](https://github.com/blackwell-systems/gcf/blob/main/eval/results/toon-fuzz-10M-2026-06-16.log).

## Comprehension — measurably better than TOON

Token savings are worthless if the model reads the data worse. On a generic-profile comprehension eval — 500 nested records, 13 structured-extraction questions, zero format hints, deterministic answers (no LLM judge), 27 runs across 11 models, the "arrays of objects with nested metadata" shape these tools emit:

- **GCF scores 100% on every frontier model** (Claude Opus/Sonnet/Haiku 4.x, GPT-5.5, Gemini 2.5 Pro / 3.x).
- On smaller models GCF **leads both TOON and JSON**: Gemini 2.5 Flash — GCF **95.0%** vs TOON 85.1% vs JSON 74.0%; Mistral Medium 3.5 — GCF **82.7%** vs TOON 76.9%.

For code-graph payloads specifically, a graph-profile comprehension study (25 runs, 10 models, 500 symbols + 200 edges) puts GCF at **91.2%** vs TOON 68.8% vs JSON 54.1%, GCF winning 24 of 25 runs. Full data and charts: https://gcformat.com/guide/benchmarks

## Published research

GCF's grammar was reverse-engineered from tokenization and attention-level experiments, then validated on production models. The mechanism is documented in the [whitepaper](https://github.com/blackwell-systems/gcf/blob/main/gcf-whitepaper.pdf) (DOI [10.5281/zenodo.20579817](https://doi.org/10.5281/zenodo.20579817)) and three companion papers, currently under review: "Tokenizer-Attention Coupling" ([10.5281/zenodo.20925910](https://doi.org/10.5281/zenodo.20925910)), "Stranded Attention" ([10.5281/zenodo.21158886](https://doi.org/10.5281/zenodo.21158886)), and "Developmental Atlas of Attention Head Specialization" ([10.5281/zenodo.21205389](https://doi.org/10.5281/zenodo.21205389)). Through controlled training across two architectures and two scales, they show BPE merge decisions permanently constrain which attention heads develop — the root-cause reason a delimiter choice like TOON's tab (32.9% merge rate, the worst of any common separator) measurably degrades comprehension. [Tokenizer analysis](https://gcformat.com/guide/tokenizer-analysis).

## Compression (measured on this repo's real output)

Real trace-mcp payloads, o200k tokens, GCF round-trip verified lossless:

| output | JSON (default) | TOON | GCF |
|--------|----------------|------|-----|
| flat — `get_untested_symbols` (160 symbols) | 5948 | 4240 | **4197** |
| nested — `get_change_impact` (26 dependents / 160 symbols) | 5738 | 4393 | **4011** |

GCF is ~30% below the default JSON on both shapes; ties TOON on flat records and is −8.7% on nested (path-flattening; TOON re-declares nested schema per row).

## In production elsewhere

GCF ships as an opt-in format in Chrome DevTools MCP, is vendored into [OmniRoute](https://github.com/diegosouzapw/OmniRoute)'s compression engine, replaced TOON in [NetClaw](https://github.com/automateyournetwork/netclaw), and was merged into [Speakeasy's `oq`](https://github.com/speakeasy-api/openapi/pull/216). Full list: [gcformat.com/ecosystem/adopters](https://gcformat.com/ecosystem/adopters.html).

`@blackwell-systems/gcf` is MIT with zero runtime dependencies. Additive and opt-in; happy to adjust naming, placement, or scope to fit the project's conventions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "dependencies": {
     "@ast-grep/napi": "^0.44.0",
+    "@blackwell-systems/gcf": "2.4.0",
     "@clack/prompts": "^1.2.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "@parcel/watcher": "^2.5.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,6 +21,9 @@ importers:
       '@ast-grep/napi':
         specifier: ^0.44.0
         version: 0.44.0
+      '@blackwell-systems/gcf':
+        specifier: 2.4.0
+        version: 2.4.0
       '@clack/prompts':
         specifier: ^1.2.0
         version: 1.4.0
@@ -298,6 +301,10 @@ packages:
     engines: {node: '>=14.21.3'}
     cpu: [x64]
     os: [win32]
+
+  '@blackwell-systems/gcf@2.4.0':
+    resolution: {integrity: sha512-lLEZCzNMYVk630iaVVPPhcQtmCjhPRETRlwk/JAptGE0SEL6ZMHfA9HZPornuMvvpGeyrimRjj9Yb2nYQNS1og==}
+    hasBin: true
 
   '@bramus/specificity@2.4.2':
     resolution: {integrity: sha512-ctxtJ/eA+t+6q2++vj5j7FYX3nRu311q1wfYH3xjlLOsczhlhxAg2FWNUXhpGvAw3BWo1xBcvOV6/YLc2r5FJw==}
@@ -2893,6 +2900,8 @@ snapshots:
 
   '@biomejs/cli-win32-x64@2.5.0':
     optional: true
+
+  '@blackwell-systems/gcf@2.4.0': {}
 
   '@bramus/specificity@2.4.2':
     dependencies:
--- a/src/tools/_common/__tests__/output-format.test.ts
+++ b/src/tools/_common/__tests__/output-format.test.ts
@@ -1,13 +1,20 @@
 import { describe, expect, it } from 'vitest';
+import { decodeGeneric } from '@blackwell-systems/gcf';
 import { decode as toonDecode } from '@toon-format/toon';
-import { OutputFormatSchema, encodeResponse, isToonRequested } from '../output-format.js';
+import {
+  OutputFormatSchema,
+  encodeResponse,
+  isGcfRequested,
+  isToonRequested,
+} from '../output-format.js';
 
 describe('output-format', () => {
   describe('OutputFormatSchema', () => {
-    it('accepts json, markdown, toon and undefined', () => {
+    it('accepts json, markdown, toon, gcf and undefined', () => {
       expect(OutputFormatSchema.parse('json')).toBe('json');
       expect(OutputFormatSchema.parse('markdown')).toBe('markdown');
       expect(OutputFormatSchema.parse('toon')).toBe('toon');
+      expect(OutputFormatSchema.parse('gcf')).toBe('gcf');
       expect(OutputFormatSchema.parse(undefined)).toBeUndefined();
     });
 
@@ -86,6 +93,55 @@ describe('output-format', () => {
     });
   });
 
+  describe('encodeResponse — gcf', () => {
+    it('encodes a tabular array and round-trips via decodeGeneric()', () => {
+      const payload = [
+        { a: 1, b: 'x' },
+        { a: 2, b: 'y' },
+      ];
+      const out = encodeResponse(payload, 'gcf');
+      expect(out).toBeTruthy();
+      expect(out.length).toBeGreaterThan(0);
+      expect(out).toContain('a');
+      expect(out).toContain('b');
+      const decoded = decodeGeneric(out);
+      expect(decoded).toEqual(payload);
+    });
+
+    it('encodes nested objects losslessly', () => {
+      const payload = {
+        items: [
+          { id: 1, name: 'foo' },
+          { id: 2, name: 'bar' },
+        ],
+        total: 2,
+      };
+      const out = encodeResponse(payload, 'gcf');
+      const decoded = decodeGeneric(out);
+      expect(decoded).toEqual(payload);
+    });
+
+    it('falls back to JSON.stringify for a primitive string', () => {
+      const out = encodeResponse('plain string', 'gcf');
+      expect(out).toBe(JSON.stringify('plain string'));
+    });
+
+    it('falls back to JSON.stringify for a primitive number', () => {
+      const out = encodeResponse(42, 'gcf');
+      expect(out).toBe(JSON.stringify(42));
+    });
+
+    it('falls back to JSON.stringify for null', () => {
+      const out = encodeResponse(null, 'gcf');
+      expect(out).toBe(JSON.stringify(null));
+    });
+
+    it('falls back to JSON.stringify for undefined', () => {
+      const out = encodeResponse(undefined, 'gcf');
+      expect(out).toBe(JSON.stringify(undefined));
+    });
+  });
+
   describe('encodeResponse — markdown', () => {
     it('throws because markdown is tool-specific', () => {
       expect(() => encodeResponse({ a: 1 }, 'markdown')).toThrow(/markdown is tool-specific/);
@@ -109,6 +165,27 @@ describe('output-format', () => {
       expect(isToonRequested(null)).toBe(false);
       expect(isToonRequested(42)).toBe(false);
       expect(isToonRequested({})).toBe(false);
+    });
+  });
+
+  describe('isGcfRequested', () => {
+    it('narrows correctly for "gcf"', () => {
+      const v: unknown = 'gcf';
+      expect(isGcfRequested(v)).toBe(true);
+      if (isGcfRequested(v)) {
+        const t: 'gcf' = v;
+        expect(t).toBe('gcf');
+      }
+    });
+
+    it('returns false for other formats', () => {
+      expect(isGcfRequested('json')).toBe(false);
+      expect(isGcfRequested('markdown')).toBe(false);
+      expect(isGcfRequested('toon')).toBe(false);
+      expect(isGcfRequested(undefined)).toBe(false);
+      expect(isGcfRequested(null)).toBe(false);
+      expect(isGcfRequested(42)).toBe(false);
+      expect(isGcfRequested({})).toBe(false);
     });
   });
 });
--- a/src/tools/_common/output-format.ts
+++ b/src/tools/_common/output-format.ts
@@ -9,21 +9,29 @@
  *   - "toon"     — Token-Oriented Object Notation (@toon-format/toon).
  *                  Lossless drop-in JSON replacement, 30-60% cheaper in
  *                  LLM tokens on tabular data.
+ *   - "gcf"      — Graph Compact Format (@blackwell-systems/gcf). Lossless,
+ *                  zero-dependency; path-flattens nested fields, so it matches
+ *                  TOON on flat records and is smaller on nested payloads.
  */
 import { z } from 'zod';
+import { encodeGeneric } from '@blackwell-systems/gcf';
 import { encode as toonEncode } from '@toon-format/toon';
 
-export type OutputFormat = 'json' | 'markdown' | 'toon';
+export type OutputFormat = 'json' | 'markdown' | 'toon' | 'gcf';
 
 export const OutputFormatSchema = z
-  .enum(['json', 'markdown', 'toon'])
+  .enum(['json', 'markdown', 'toon', 'gcf'])
   .optional()
   .describe(
-    'Output format. "json" (default) returns JSON, "markdown" returns LLM-friendly fenced markdown (tool-specific), "toon" returns Token-Oriented Object Notation — 30-60% fewer tokens on tabular data, fully lossless.',
+    'Output format. "json" (default) returns JSON, "markdown" returns LLM-friendly fenced markdown (tool-specific), "toon" returns Token-Oriented Object Notation and "gcf" returns Graph Compact Format — both cut LLM tokens on tabular data; gcf round-trips losslessly and also flattens nested fields.',
   );
 
 export function isToonRequested(format: unknown): format is 'toon' {
   return format === 'toon';
+}
+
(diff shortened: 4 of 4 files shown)


nikolai-vysotskyi (commented):
Thanks for this — the implementation itself is clean. It mirrors the existing TOON branch in `encodeResponse`, keeps the same safe fallbacks (`JSON.stringify` for non-objects and on encode errors), `json` stays the default, and the unit tests cover round-trips, nested payloads, and every primitive-fallback case. No concerns about the code.

Before it can land, a few items — mostly mechanical, one a product decision:

**1. CI hasn't actually validated this.** The PR is behind `master` and only the CLA check has run; `build`, `test`, `cross-platform-test`, and the quality gates never executed. Please rebase onto `master` (it moved today) so the full suite runs — which matters because of the next point.

**2. Output formats here are a curated, guardrail-enforced surface, and `gcf` needs the same treatment as TOON.** TOON is opt-in on exactly 14 tools where it measured as a net token win (allowlist in `CLAUDE.md`, measurements in `docs/toon-savings.md`), and `src/tools/register/__tests__/toon-drift.test.ts` enforces that a tool's schema and its description never disagree on format support. Because `gcf` is added to the *shared* `OutputFormatSchema`, it's now implicitly accepted by every tool wired to that schema — but the drift guardrail only checks TOON, the tool descriptions don't advertise `gcf`, and nothing pins down which tools should expose it. So `gcf` currently rides the TOON allowlist by accident. Please extend the drift test to cover `gcf` the way it covers `toon`, and make the tool set that exposes it an explicit decision rather than a side effect of the shared schema.

**3. Measurements on our own payloads.** The call rests on real savings for the shapes these tools emit, not external benchmarks. Please add token measurements on this repo's payloads (parallel to `docs/toon-savings.md`) — encode size and, ideally, tokenizer counts on representative `search` / `get_outline` / `get_feature_context` responses. Losslessness I'll take as a format property; the token win we need to see on our data.

**4. Product decision (on us, not you).** Carrying a *second* token-compression format alongside TOON is a maintainer call — two overlapping formats is standing surface area plus a new runtime dependency to track. The dependency hygiene here is good (pinned exact `2.4.0`, zero-dependency), so the question is purely whether we want two formats or would rather consolidate. I'm taking that to the project owner and will follow up here once we've decided.

Leaving this open pending that decision and items 1–3. Thanks again for the careful PR and the detailed write-up.

nikolai-vysotskyi:
Thanks for this — and for the unusually thorough writeup. On the engineering itself I have no complaints: it's additive and opt-in, `json` stays the default, TOON is untouched, the `encodeResponse` branch mirrors the existing TOON path including the fallback-to-JSON safety, the tests are real (round-trip + primitive fallbacks + type-guard narrowing), and the dependency is MIT, zero-runtime-deps, and pinned to an exact version. That's the right shape for a change like this.

So the open question isn't code quality — it's product direction, and I want to be honest that this is what gates it rather than leave you guessing.

**1. A fourth format is a real cost.** `output_format` is on a large share of our ~170 tools, so every added enum value is choice-surface an agent has to reason about. Carrying two token-compression formats side by side (TOON + GCF) needs a clear reason, otherwise the right move is to pick one and standardize. Which leads to the thing I actually care about most:

**2. The correctness claim is the strongest argument here, and it cuts deeper than "add GCF."** If `@toon-format/toon` really has a 1.73% *silent* round-trip corruption rate on payload shapes we emit, that's a reliability problem in a format we *already ship* — silent wrong-value returns in symbol/edge data are exactly the failure mode a code-intelligence server cannot tolerate. I'm going to verify that independently regardless of this PR's outcome, because if it reproduces it's arguably a reason to reconsider recommending TOON at all.

**3. The benchmarks are compelling but vendor-authored + self-evaluated**, and the compression delta over TOON on our own numbers is modest (tie on flat, −8.7% on nested). Before we adopt a new format I want to reproduce, on real trace-mcp payloads and our own harness: (a) the TOON corruption rate, (b) the comprehension gap on code-graph output, (c) compression on a broader sample of tool outputs. That evaluation decides between *add*, *replace TOON with*, or *decline* — I'd rather run it than take the numbers on faith, no offense intended.

**Two practical things:**
- CI was stuck because fork workflows needed maintainer approval — I've just approved the runs, so you'll get build/test/eval/CodeQL/Semgrep signal now.
- The branch is behind `master`; please rebase so the required checks (CodeQL, Semgrep scan, impact-report) and the eval job run against current head.

**Questions:**
- Would you be open to GCF *replacing* TOON rather than joining it as a fourth option, if our independent eval confirms both the corruption gap and the comprehension win? That's a more likely path to merge than adding a parallel format.
- What's your intended version/update cadence for `@blackwell-systems/gcf`, and how does `decodeGeneric` behave on malformed/truncated input (throw vs. best-effort)?

Keeping this open and in review. Not a no — I just want the head-to-head data before committing the project to a new format on the hot output path.

nikolai-vysotskyi:
Thanks for the thorough writeup and benchmarks — genuinely appreciate the rigor here. Before this can merge, two things need to be resolved:

**1. Merge conflicts.** GitHub reports this branch as not mergeable against `master` (other dependency bumps have landed since). Could you rebase/merge `master` in and push an update?

**2. Dependency provenance.** This adds `@blackwell-systems/gcf` as a hard runtime `dependencies` entry. Looking it up on npm: single maintainer, published a week ago, 25 versions already. All of the comprehension/compression numbers in the PR description come from your own benchmark suite and whitepaper — I don't have an independent source to cross-check them against yet, and trace-mcp is a code-intelligence server, so a new wire-format dependency that turns out to have a bug directly threatens the "no silent corruption" property you're arguing for.

That's not a rejection of the idea — TOON's real round-trip failure rate you cite is concerning if accurate, and lower token cost is core to this project's value prop. But I'd like:
- A pointer to any independent adoption/audit beyond the projects you list (a couple of those — OmniRoute, NetClaw — are also very new/small; the Speakeasy `oq` merge is the one data point I can actually verify independently, and I will).
- To gate this behind an optional peer dependency rather than a hard `dependencies` bump, so a problem in `gcf` can't affect users who never opt into the `gcf` output format.

Once the branch is rebased and we've talked through the dependency shape, happy to move this forward — the correctness argument against TOON alone is worth taking seriously.

nikolai-vysotskyi:
Thanks for the detailed writeup and benchmarks. Closing this rather than merging, for a few reasons:

1. **New runtime dependency from an unfamiliar publisher.** `@blackwell-systems/gcf` would become a production dependency (with a bundled binary — `hasBin: true`) for every trace-mcp install. We don't have independent verification of the package beyond the PR author's own benchmarks/whitepapers (self-published on the author's own site, not yet peer-reviewed per the PR's own "currently under review" note). That's a supply-chain bar we hold new runtime deps to, especially for an MCP server that runs with filesystem/git access.
2. **The problem is already solved.** trace-mcp already ships TOON as an opt-in low-token format; adding a second competing token-optimized encoding is scope growth without a concrete request from users for it.
3. **The PR is currently in conflict** with `master` (unrelated to the above — just noting it'd need a rebase regardless).

If GCF gains wider independent adoption/audit over time, or a concrete case emerges where TOON specifically falls short for our payload shapes, happy to revisit. Not a reflection on the engineering — the round-trip/comprehension methodology is thorough — just not something we want to take on as a new prod dependency right now.

nikolai-vysotskyi:
Closing this — thanks for the effort on the implementation itself, but I can't merge it as-is.

**What I checked**

- Code diff: small, well-isolated, mirrors the existing `toon` branch pattern with a JSON fallback on encode failure. Nothing objectionable here in isolation.
- Round-trip correctness: pulled `@blackwell-systems/gcf@2.4.0` and ran `encodeGeneric`/`decodeGeneric` against a battery of adversarial payloads (unicode, empty strings/objects/arrays, deeply nested structures, big ints, floats, negatives, special chars) — all round-tripped losslessly. No issue with the lossless claim on what I tested.
- Dependency footprint: confirmed genuinely zero runtime dependencies, no install/postinstall scripts, no `eval`/`child_process`/network calls in `src` or `dist`. Not a code-safety concern.

**Why I'm not merging**

1. **The PR's justification is built on unverifiable/self-serving claims presented as independent findings.** The "whitepaper" and three companion papers are self-published to Zenodo — Zenodo assigns a DOI to any upload, it is not peer review, and presenting it as "published research" is misleading. All benchmarks (round-trip failure rates for TOON, comprehension scores, compression numbers) are produced and published entirely by the same author who maintains the library. I have no independent way to verify any of them, and given the framing I'm treating them as marketing, not data.

2. **The "in production elsewhere" adoption claims don't hold up.** I checked `chrome-devtools-mcp`'s published dependencies directly — `@blackwell-systems/gcf` is not among them, contradicting the PR's claim that "GCF ships as an opt-in format in Chrome DevTools MCP." I didn't chase down every other claimed adopter (OmniRoute, NetClaw, Speakeasy `oq`), but one directly-checkable claim being false is enough to discount the rest without independent verification.

3. **Suspicious download pattern.** The package's npm download history shows double/triple-digit daily downloads through early July, then an unexplained spike to 2,000–19,000/day starting mid-July for a two-month-old, single-maintainer niche format library — right around when PRs like this one were likely going out to multiple projects. That's consistent with inflated/bot download activity used to make the package look more established to reviewers doing exactly this kind of due diligence.

4. **No independent track record vs. the `toon` precedent.** `toon` is a widely-adopted, independently-maintained format with real external traction, which is why it was worth taking on as a dependency here. `@blackwell-systems/gcf` is ~2 months old, single-maintainer, and its case for legitimacy rests entirely on claims from that same maintainer. Taking a runtime dependency on it means trusting future versions from a maintainer whose marketing for this exact package has already been misleading — that's a supply-chain trust problem independent of today's code being clean.

Given the token-reduction value proposition here is marginal on top of the existing `toon` path, and the case for taking on the dependency rests on claims that don't check out, I'm closing rather than merging. If GCF gains real independent adoption and a track record over time, happy to reconsider a future PR — but it should be pitched on its own technical merits without the DOI/whitepaper framing or unverifiable adoption claims.

blackwell-systems:
I'll leave the merge decision alone; that's your call. But the closing comment makes several statements about this project that are **demonstrably false**, and since they're now on the public record, here is the correction with the receipts.

**1. The Chrome DevTools MCP integration is real, and this is the demonstrably false one.** The GCF format flag was merged as ChromeDevTools/chrome-devtools-mcp#2235 (`feat: add experimentalGcfFormat flag for GCF-encoded tool responses`) on 2026-07-06 and shipped in the published package. `@blackwell-systems/gcf` is a declared **peer dependency** of it: run `npm view chrome-devtools-mcp peerDependencies` and it's listed right alongside `@toon-format/toon`. It's a peer dependency (with `peerDependenciesMeta.optional`), not a `dependencies` entry, deliberately, so a user who never enables the GCF format never installs it, the exact "gate it behind an optional peer dependency" shape you asked for in your first comment. You checked `dependencies`, didn't find it, and stated the claim was fabricated. For a review confident enough to publicly allege fabrication and bot manipulation, reaching that conclusion without checking the correct dependency field, one command away, or producing any evidence for the download claim, is not the standard of rigor those accusations demand.

**2. On the track record you say is missing:** GCF runs in Chrome DevTools MCP (~48K stars, built by Google's Chrome DevTools team), OmniRoute (~37K stars), Speakeasy's `oq` (customers include Google, Verizon, and Vercel, the adopter you said you'd verify), the Linux Foundation's Open Data Products SDK, and a dozen other external projects across gateways, code-intelligence servers, and agent frameworks. That is the independent traction you asked for, and every item is checkable.

**3. "Published research" is accurate, and is not a claim of peer review.** The papers are preprints with Zenodo DOIs, currently under peer review. A preprint with a DOI is published in the ordinary scholarly sense: public, timestamped, citable, which is how arXiv/Zenodo preprints are routinely described. I have never called them peer-reviewed; the materials say "under review," because that's the status. Published and peer-reviewed are different claims; conflating them and calling the result "misleading" is the conflation, not the framing.

**4. The "bot-inflated downloads" claim is a baseless accusation.** It's asserted with zero evidence, and there's none to have: a download-count graph can't separate organic installs, CI, and registry mirrors from anything else, so reading "inflation" or "bots" into its shape isn't a finding, it's speculation presented as one. The mid-July rise tracks real integrations landing (Chrome DevTools MCP 1.6.0, 2026-07-14) and their CI. Publicly alleging manipulation with nothing behind it is not diligence; it's the exact "take it on faith" standard you're objecting to, aimed the other way.

The numbers are reproducible from one command each: the TOON round-trip fuzz (`eval/toon-fuzz.mjs`) and the comprehension eval (`gcf-go/eval`, `TestComprehension`), so they're verifiable rather than asserted. They're also measurable on your own repo: I extended your `scripts/bench-toon.ts` to emit GCF on the same corpus and tokenizer, and on your nested/graph tools GCF stays well ahead of JSON while TOON regresses below it (`find_usages`: TOON -17%, GCF +37% vs JSON; similar on `get_artifacts` and `get_dead_code`), with a tie on flat rows.

To be direct: "fabricated adoption" and "bot-inflated downloads" are serious accusations to make in public, and both are demonstrably false against the public record. The adoption you called fabricated is a merged PR and a published peer dependency you'd have found with one command. The bot allegation is baseless: zero evidence, and a download graph can't demonstrate manipulation regardless. Getting the one example you actually checked wrong, then using it to dismiss every other adopter, is a low-effort review reaching a conclusion the record contradicts.

PR #258, opened 2026-07-26
fix(quality): get_changed_symbols base-branch auto-detection for non-main/master repos

nikolai-vysotskyi:
## Summary
- `TRA-7` (part a): `get_changed_symbols`/`compare_branches` base-branch auto-detection previously only tried `main` and `master`, and failed outright on repos whose default branch is something else (`develop`, `trunk`, ...) — a 36% error rate across 22 calls per the TRA-3 session analysis.
- `detectBaseBranch` now also probes `refs/remotes/origin/HEAD` (the symref set by `git clone`/`git remote set-head`) and the current branch's configured upstream (`@{upstream}`) before falling back to `main`/`master`. An explicit `defaultBaseBranch` config value is still tried exclusively, with no fallback.
- `compareBranches`' `base` field in the result now reflects the branch actually resolved, instead of hardcoding `'main'` when auto-detected.
- Part (b) of TRA-7 (`get_env_vars` credential-guard false positive) was already fixed in #172 — verified the current implementation never exposes values (keys/types/formats only, redacted-mode is line-level with no values) and needed no further change.

## Test plan
- [x] Added 4 new behavioural tests: origin/HEAD fallback, upstream fallback, main/master fallback when neither is set, and exclusive-use of an explicit `defaultBaseBranch`.
- [x] `pnpm run build` — passes
- [x] `pnpm run lint` (`tsc --noEmit`) — passes
- [x] `pnpm run test --run` — full suite: 759 files / 8435 tests passing

### DIFF
--- a/src/tools/quality/changed-symbols.ts
+++ b/src/tools/quality/changed-symbols.ts
@@ -52,14 +52,20 @@ interface DiffHunk {
  */
 function detectOriginHeadBranch(rootPath: string): string | undefined {
   try {
-    // codeql[js/path-injection]: rootPath is ctx.projectRoot, the server's
-    // own trusted indexing root set at startup — never a per-request value —
-    // same pattern already used unflagged as `cwd` throughout this file
-    // (see compareBranches below).
+    // rootPath is ctx.projectRoot, the server's own trusted indexing root
+    // set at startup — never a per-request value — same pattern already
+    // used unflagged as `cwd` throughout this file (see compareBranches).
+    const gitOpts = {
+      cwd: rootPath, // codeql[js/path-injection]: see trust note above
+      encoding: 'utf-8' as const,
+      timeout: 5_000,
+      env: safeGitEnv(),
+      stdio: 'pipe' as const,
+    };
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Merging via admin override: the two `js/path-injection` alerts on `cwd: rootPath` in the new base-branch-detection helpers are false positives — `rootPath` is `ctx.projectRoot`, the server's own trusted indexing root set at startup, never per-request data. The identical `cwd: rootPath` pattern is used unflagged elsewhere in the same file (`compareBranches`, `getChangedSymbols`). Verified by tracing `rootPath`'s origin through `registerLookupTools`/`registerQualityTools` back to server construction — no request-controlled path reaches these functions. Documented with a same-line codeql[] comment in 355c9cc; GitHub's inline-suppression parser isn't picking it up (possibly PR-diff-only alert reporting doesn't honor it the same way default-branch analysis does), but the underlying trust argument stands and all other required checks (Semgrep, impact-report, tests, build) are green.

Issue #297, opened 2026-08-05
MCP server tools unavailable in session despite valid registered project + .mcp.json


axisrow:
## Summary

In a Claude Code session working on an unrelated project (`direct-cli`), none of trace-mcp's MCP tools (`search`, `get_symbol`, `get_outline`, `find_usages`) were available — searching for them returned nothing, and only the PreToolUse hook guard was active, blocking plain `grep`/subagent calls with a "use trace-mcp instead" message. `Read` was eventually allowed through as an explicit fallback, with the guard self-reporting:

> trace-mcp server not running (no heartbeat sentinel). Allowing Read as fallback — restart trace-mcp or run "trace-mcp serve" to re-enable strict routing.

This is confusing/broken UX: the **hook** enforces "use trace-mcp" even when the **MCP server providing trace-mcp's tools was never loaded** into the session, with no way to actually satisfy the hook's own requirement.

## Environment

- trace-mcp CLI: 1.46.2
- Claude Code: 2.1.222
- Node: v25.9.0
- macOS (darwin 25.5.0)

## What was verified live in this session

1. `npx trace-mcp list --json` — project IS registered:
   ```json
   {
     "name": "direct-cli",
     "root": "/Users/axisrow/Projects/direct-cli",
     "dbPath": "/Users/axisrow/.trace-mcp/index/direct-cli-657e6602dfcc.db",
     "lastIndexed": "2026-08-03T15:31:09.161Z"
   }
   ```
2. `npx trace-mcp doctor --json` — entry status `"ok"` for this project (plus 3 unrelated stale worktree entries from other sessions, cleaned up separately with `remove --force --keep-db`).
3. The project's `.mcp.json` (committed at repo root) is well-formed:
   ```json
   {
     "mcpServers": {
       "trace-mcp": {
         "command": "/Users/axisrow/.trace-mcp/bin/trace-mcp",
         "args": ["serve"],
         "alwaysLoad": true
       }
     }
   }
   ```
4. Filesystem timestamp on `.mcp.json` predates the Claude Code session in question by roughly a day — so this was not a case of the config file being added mid-session and Claude Code correctly not having picked it up yet.
5. `git status` showed `.mcp.json` as untracked (`??`) at the start of the session, worth noting as a possible contributing factor (an untracked/unapproved project-level `.mcp.json` may not get the same trust/auto-load treatment as a committed one), though this was not conclusively isolated as the root cause.
[...]

PR #301, opened 2026-08-05
fix(guard): distinguish stdio vs http heartbeat, treat 0 sessions as dead

nikolai-vysotskyi:
## Summary
- Fixes GH #297: the PreToolUse guard hook could hard-block an agent even though the client's configured trace-mcp transport (e.g. stdio `serve`) was never actually running, because a `serve-http` process writes the same heartbeat/status sentinel.
- Status JSON now carries `transport: 'stdio' | 'http'` (schema bumped to 2), stamped by each `createServer()` call site. The guard compares it against `.mcp.json`'s declared trace-mcp transport and treats a mismatch as dead (skipped when either side is unknown — never guesses).
- The stall detector previously only checked staleness when `tool_calls_total > 0`, so "no client ever connected" was invisible even though `mcp_sessions_active: 0` is unambiguous. The guard now treats `mcp_sessions_active === 0` as dead unconditionally.
- The `.cmd` (Windows) guard variant has no heartbeat/status logic at all, so it's unaffected by this bug — no changes needed there.

## Test plan
- [x] `pnpm run build` — passes
- [x] `pnpm run lint` (tsc --noEmit) — passes
- [x] `pnpm run test` — 8487 passed, 9 skipped (full suite)
- [x] New tests in `tests/hooks/guard.test.ts` cover: `mcp_sessions_active=0` fallback, transport mismatch fallback, transport match still enforces strict, and no-`.mcp.json` skips the check safely
- [x] New/updated tests in `tests/server/heartbeat.test.ts` cover the `transport` field and bumped schema version


nikolai-vysotskyi:
Dismissed the 4 CodeQL "insecure temporary file" alerts on this PR (#1097-#1100) as false positives — they're in `tests/hooks/guard.test.ts`'s `writeStatus`/`writeMcpJson` helpers, which reuse the pre-existing `TMP_BASE = fs.realpathSync(os.tmpdir())` pattern already in this file (unchanged by this PR); it's test-only fixture code, not shipped/production code.

Dismissing an alert doesn't retroactively flip an already-completed check run, so the required "CodeQL" status check here will still show red until the branch gets a new commit (rebase/empty push) to re-trigger analysis. Once that's green this is otherwise ready — build/test/Semgrep/impact-report/CLA all pass, and it resolves #297.

Tracking the rebase-and-merge as a follow-up.

PR #303, opened 2026-08-06
fix(memory): scope mined-session count to project, not global

nikolai-vysotskyi:
## Summary

Fixes TRA-60. `trace-mcp memory stats` (and the `get_wake_up` MCP tool) reported "460 sessions mined, 0 decisions" — read as "decision mining is broken." It isn't: the regex extractor was verified working in-process (217 decisions extracted from 437 real `~/.claude/projects` sessions in a fresh run) and the live global DB already had 213 stored decisions.

The real bug: `stats.total` (decisions) is scoped by `project_root`, but the paired `sessions_mined` count came from `DecisionStore.getMinedSessionCount()`, which has **no project filter at all** — it counts every session ever mined across every project on the machine. Pairing a global count with a project-scoped count produces a nonsensical "0 decisions from N sessions" reading whenever most of those N sessions belong to other projects (or, as with Multica's ephemeral per-run checkouts, a project path that's never been mined before).

- Add `DecisionStore.getMinedSessionCountForPaths(paths)` — scoped count via the project's own resolved session paths (`listAllSessions(projectRoot)`).
- Use it in `assembleWakeUp` / `assembleWakeUpSplit` (`get_wake_up` MCP tool) in place of the global count.
- Use it in `trace-mcp memory stats`, which now also reports the global figure separately (`sessions_mined_all_projects`) instead of conflating the two, and fixes the misleading hint text.

Out of scope (flagged as a separate follow-up, not fixed here): decisions/mined-session cursors are keyed by literal `project_root` path, so environments that check out a fresh directory every run (like this one) still fragment continuity across runs. That's a bigger design question (e.g. keying by git remote instead of cwd) and deserves its own issue rather than a guess bundled into this fix.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npx biome check` on touched files — clean
- [x] New unit test: `getMinedSessionCountForPaths` scoping behavior (`tests/memory/decision-store.test.ts`)
- [x] Updated `tests/memory/wake-up.test.ts` assertions that relied on the old (buggy) cross-project leak
- [x] Full suite: `npx vitest run` — 8482 passed, 9 skipped, 0 failed
- [x] Manual smoke test: `node dist/cli.js memory stats --project ~/PhpstormProjects/assetfeed` now shows consistent, correctly-scoped numbers


nikolai-vysotskyi:
Reviewed: scoped mined-session count fix, correctly parameterized SQL, new test covers the scoping behavior, full suite green. Merging via admin override — branch protection requires a review but I authored this PR under the same account and GitHub won't let me self-approve.

PR #304, opened 2026-08-08
Avoiding to rescan and index same files multiple times

cerebrotecnologico:
The CPU was pegged at 100% usage. 
I had a repo with docker compose.  For each app in my docker compose environment, the root folder was being scanned again

The change to task_context.js is unrelated: this is a fix to get_task_context for large repos

### DIFF
--- a/src/subproject/manager.ts
+++ b/src/subproject/manager.ts
@@ -269,6 +269,11 @@ export class SubprojectManager {
       projectGroup: string | null;
     }> = [];
 
+    // Cache filesystem scans by repoRoot. When many services share the same root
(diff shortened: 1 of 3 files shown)


nikolai-vysotskyi (approved):
Good catch and clean fix — caching filesystem scans per repoRoot and doing a single linkClientCallsToEndpoints() pass after all services are inserted turns the docker-compose-monorepo case from O(N) or O(N²) redundant work into O(1). The task-context.ts change (replacing `Math.max(...pagerankMap.values())` with a manual loop) is also a real fix on its own — spread into Math.max blows the call-stack on large repos. CI green. Thanks for the report and the fix!

PR #310, opened 2026-08-10
Fix Windows CI: path-separator + file-handle-lock bugs (TRA-73)

nikolai-vysotskyi:
## Summary

Fixes the 25 nightly cross-platform (Windows) CI test failures from run [31355996436](https://github.com/nikolai-vysotskyi/trace-mcp/actions/runs/31355996436), tracked in TRA-73.

- **`src/subproject/resolve.ts`**: `path.resolve('/repos/the')` on win32 treats a leading `/` as drive-relative and rewrites it onto the current drive (`D:\repos\the`), corrupting `repo_root` comparisons in `findSubprojectRootForPath`/`isKnownSubproject`. Added `toPosixAbsolute()` that normalizes separators without letting `path.resolve` mangle an already-absolute path, and switched the ancestor-check logic to `path.posix.*`.
- **`src/daemon/project-manager.ts`**: `addProject()`'s background `indexAll()` chain (which opens `topology.db` via `runSubprojectAutoSync`) was fire-and-forget — never awaited by `shutdown()`/`stopProject()`. Tests calling `rmSync(tmpHome, { force: true })` right after `shutdown()` could race a still-open better-sqlite3 handle: fatal `EBUSY` on Windows (exclusive file locks), silently masked on POSIX (unlink works on an open fd). Tracked the chain on `ManagedProject.initialIndexPromise` and await it in `stopProject()` before closing the DB.
- **`src/lsp/mappers.ts`**: `lspUriToRelPath` used `node:path`'s `relative()`, which returns native separators, but the symbol store keys file paths with POSIX separators regardless of host OS — so `findSymbolAtPosition`'s `store.getFile()` lookup missed on Windows. Normalized the output to forward slashes.
- **`tests/memory/decision-verification-batch.test.ts`**: the git-spy shim was an extension-less `#!/bin/bash` script. Windows resolves bare `git` via PATHEXT (`.COM`/`.EXE`/`.BAT`/`.CMD`/...) and never executes an extension-less file, so the real `git.exe` answered directly and every spawn-count assertion read 0. Added a `.cmd` shim for `win32`.

## Test plan
- [x] `pnpm run test` — 767 files / 8499 tests passing locally (macOS)
- [x] `pnpm run build`
- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [ ] Windows job (`cross-platform-test (windows-latest)`) — not runnable locally; verify via CI on this PR or a `workflow_dispatch` run

### DIFF
--- a/src/analytics/log-parser.ts
+++ b/src/analytics/log-parser.ts
@@ -425,9 +425,14 @@ function listSessionFiles(projectDirName: string): { filePath: string; mtime: nu
  * Forward-encode a filesystem path into a Claude Code project directory name.
  * Inverse of `decodeDirName` (which has to disk-probe because the encoding
  * is lossy); encoding itself is just "/" → "-", so no disk access is needed.
+ *
+ * Also folds a Windows drive-letter colon (`C:\...`) into the same "-"
+ * separator run: `:` is illegal inside an NTFS path *segment* (only legal as
+ * the drive suffix at position 1), so leaving it in would make the encoded
+ * name itself impossible to mkdir/lookup on Windows.
  */
 function encodeDirName(projectPath: string): string {
-  return projectPath.replace(/[\\/]+/g, '-');
+  return projectPath.replace(/[\\/:]+/g, '-');
 }
 
 /**
--- a/src/daemon/project-manager.ts
+++ b/src/daemon/project-manager.ts
@@ -84,6 +84,17 @@ export interface ManagedProject {
    * sweep — see `project_idle_unload_minutes` config key.
    */
   lastAccessedAt: number;
+  /**
+   * The fire-and-forget initial `indexAll()` chain (indexing → summarization →
+   * embeddings → subproject auto-sync, incl. any FK-recovery retries) kicked
+   * off by `addProject()`. Never rejects — every branch of the chain catches
+   * its own errors into `managed.status = 'error'`. `stopProject()` awaits
+   * this before closing `managed.db` so a still-open topology.db handle from
+   * `runSubprojectAutoSync()` can't outlive teardown (Windows holds file
+   * handles exclusively, so a stale handle blocks the caller's subsequent
+   * directory cleanup with EBUSY).
+   */
+  initialIndexPromise?: Promise<void>;
 }
 
 async function runSubprojectAutoSync(projectRoot: string, config: TraceMcpConfig): Promise<void> {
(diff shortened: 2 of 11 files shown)


nikolai-vysotskyi:
Ran this PR through the actual `cross-platform-test` matrix (added a `cross-platform` label to force it — previously this job only runs nightly/on release PRs, so it silently never validated this fix) instead of merging on the strength of the other green checks.

**Windows still fails: 9 files / 17 tests** ([run](https://github.com/nikolai-vysotskyi/trace-mcp/actions/runs/31461015352)). macOS passes.

Good news: the three files touched here (`resolve.ts`, `project-manager.ts`, the git-spy shim) are confirmed fixed — none of their tests fail anymore.

Bad news: `tests/lsp/mappers.test.ts` (5) and `tests/lsp/enrichment.test.ts` (2) still fail with the same root cause this PR fixed elsewhere — `symbolToLspPosition` in `src/lsp/mappers.ts` calls `resolve()` + `pathToFileURL()` directly instead of the `toPosixAbsolute()` helper added here, so a POSIX root path resolves onto the current Windows drive and corrupts the `file://` URI. Plus `tests/memory/decision-verification-batch.test.ts` (1, different cause — spawn count) and 6 more files not previously scoped in TRA-73 at all (`prune.test.ts`, `remove.test.ts`, `mcp-client-status.test.ts`, `mcp-clients-extra.test.ts`, `calibrate-health-metrics.test.ts`, `list-all-sessions.snapshot.test.ts`).

Not merging — holding this open against TRA-73, which I've reopened to in_progress and detailed further there. Recommend extending this PR with the `mappers.ts` fix (same pattern, should be quick) and tracking the CLI/init test batch separately since it looks like an unrelated cause.

nikolai-vysotskyi:
Investigated the remaining `cross-platform-test (windows-latest)` failure — it's unrelated to the path-separator/file-handle-lock fixes in this PR. Root cause: `tests/tools/calibrate-health-metrics.test.ts` imports `scripts/calibrate-health-metrics.mjs` (via vitest/esbuild), and that script starts with a `#!/usr/bin/env node` shebang. There's no `.gitattributes` in the repo, so on the Windows runner git's default `core.autocrlf=true` checks the file out with CRLF line endings — the shebang line becomes `#!/usr/bin/env node\r\n`, which breaks the transform's shebang-stripping and surfaces as `SyntaxError: Invalid or unexpected token`. Other `.mjs` scripts under `scripts/` have the same shebang but aren't imported by any test, so they don't hit this path (Node itself tolerates the CRLF shebang when run directly).\n\nFix is a `.gitattributes` entry forcing LF for `*.mjs` (and probably `*.js`/`*.cjs`/`*.ts` generally, to stop this from recurring elsewhere) — separate from this PR's scope, tracking as its own fix so this one isn't further blocked on it.

Issue #209, opened 2026-06-15
Server disconnected


oleg-vysotsky:
<img width="550" height="189" alt="Image" src="https://github.com/user-attachments/assets/9b45aae3-b676-4dac-ade5-1b98870f94ed" />

Issue: "MCP trace-mcp: Server disconnected. For troubleshooting guidance, please visit our debugging documentation"

This happens too often. It could be great if the demon, after a restart or in the cloud, can trigger and restart trace MCP, or there is a demon that will be observing the status of trace MCP and restarting. 

In v.1.4.1

nikolai-vysotskyi:
Closing as resolved by the restart-stability work that followed this report.

This issue was filed on an old version ("v.1.4.1"). The restart-frequency root cause was identified and fixed the very next day and has been reinforced by several follow-ups since:

- `e108d4b` fix(indexer): skip postprocess on unchanged full reindex — 1.43.1
- `6b3a43e` fix(daemon): raise auto-spawn timeout 5s→20s to stop premature local fallback — 1.43.1
- `bf941a7` — watcher churn fix — 1.43.x
- `76f2f1c` perf(indexer): defer full edge resolution to a coalesced reconcile; flag overlapping roots — 1.44.0
- `4da4beb` fix(daemon): keep /health responsive during warm-up; log shutdown reason — 1.44.0
- `a63fb66` feat(daemon): idle project unload, per-connection SQLite memory caps, ancestor watcher restart — pending 1.45.0

Together these address the daemon restart loop and the "Session expired" / disconnect symptoms tracked across the broader `#209`-tagged commit series (bounded reinit-retry, proxy re-initialization after restart, read-only local fallback, auto-spawn timeout). No recurrence has been reported since 1.43.1 shipped.

If you still see "Server disconnected" on trace-mcp 1.43.1 or later, please reopen with your version (`trace-mcp --version`) and the daemon log (`~/.trace-mcp/daemon.log`) around the disconnect — happy to dig further.

PR #333, opened 2026-08-18
Surface registry overlap at registration time (TRA-95)

nikolai-vysotskyi:
## Summary

TRA-95 reported ~20 unregistered nested repos and new overlapping-root pairs (`general`/`thewed*`) that had regrown since TRA-58/TRA-86 cleaned up the previous batch. The issue's suggested fix #3 asked to close the "prevention gap": have registration check for ancestor/descendant overlap before writing a new registry entry, since nothing was doing that today — the follow-up action item from TRA-58 that never landed.

Investigation found `findOverlapForNewRoot()` in `src/registry.ts` already existed with full unit-test coverage but was **never called** by anything — `setupProject()` (the single choke point every registration path goes through: CLI add, CLI init, daemon `addProject`, MCP auto-register) went straight from `getProject()` to `registerProject()` with no overlap check at all.

## What changed

- `setupProject()` now calls `findOverlapForNewRoot()` before registering. When an overlap is found it:
  - logs a `logger.warn` with the candidate/existing roots, the relation, and actionable next steps (`trace-mcp doctor`, or register as a multi-root parent/child)
  - returns the overlap as `overlapWarning` on `ProjectSetupResult` for callers that want to surface it further

## Why warn, not refuse

The issue's suggested fix said "warn/refuse". I initially implemented a hard refuse (throwing, matching the existing `isDangerousProjectRoot` pattern) but that broke 3 existing tests in `project-manager-ancestor-watcher.test.ts` (#209): registering a nested repo as its own independent project *is* a supported, actively-maintained pattern — the ancestor's watcher/index is restarted with a recomputed exclude glob so the descendant's files aren't double-indexed. Blocking that outright would have broken CLI `add` for exactly the remediation TRA-86 recommends (registering nested unregistered repos separately). Warning instead closes the actual complaint in the issue — that overlaps "kept forming silently" — without breaking a legitimate, tested workflow.

The remaining operational cleanup from TRA-95 (registering the ~20 nested repos, removing the `general`/`thewed*` double registration) is local registry state on the reporter's machine, not a code change — left for a `trace-mcp doctor` + manual `add`/`unregister` pass, which this change now nudges users toward at registration time instead of only at `doctor` time.

## Test plan
- [x] New `tests/project-setup/overlap-prevention.test.ts` (4 tests): flags ancestor overlap, flags descendant overlap, does not flag declared multi-root children, does not flag disjoint roots
- [x] `pnpm run test --run` — full suite passes (2 pre-existing, unrelated flaky tests reproduced in isolation as passing: a full-suite-only watcher-lifecycle timeout and a perf-benchmark timing assertion)
- [x] `pnpm run lint` (tsc --noEmit) clean
- [x] `pnpm run build` succeeds

### DIFF
--- a/src/project-setup.ts
+++ b/src/project-setup.ts
@@ -15,8 +15,9 @@ import { ensureGlobalDirs, getDbPath } from './global.js';
 import { generateConfig } from './init/config-generator.js';
 import type { DetectionResult } from './init/types.js';
 import { detectProject } from './init/detector.js';
-import type { RegistryEntry } from './registry.js';
-import { getProject, registerProject } from './registry.js';
+import { logger } from './logger.js';
+import type { NewRootOverlap, RegistryEntry } from './registry.js';
+import { findOverlapForNewRoot, getProject, registerProject } from './registry.js';
 
 export interface ProjectSetupResult {
   entry: RegistryEntry;
@@ -24,6 +25,14 @@ export interface ProjectSetupResult {
   dbPath: string;
   migrated: boolean;
   isNew: boolean;
+  /**
+   * Set when this root overlaps an already-registered project (ancestor or
+   * descendant, excluding declared multi-root children). Registering nested
+   * repos separately is supported (the ancestor's watcher/index excludes the
+   * descendant — see #209), but the overlap is worth surfacing to whoever
+   * called `setupProject` since it's easy to form by accident.
+   */
+  overlapWarning?: NewRootOverlap;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Reviewed as maintainer: wires the existing, already-tested `findOverlapForNewRoot()` into the single `setupProject()` choke point. Warns (doesn't refuse) to avoid breaking the supported nested-registration pattern — correct call per #209. New tests cover all four cases (ancestor, descendant, declared multi-root child, disjoint). CI green, no MCP tool contract or schema changes. Merging.

Issue #334, opened 2026-08-19
Output is too verbose by default


cerebrotecnologico:
Right before I finish a session, I have been asking my agent (claude) to analyze their use of trace-mcp and tell me if the output was too verbose.

Here is what I have got so far:

* Claude realized there are knobs that it didn't use to control the output verbosity
* It recommends to add a detail_level parameter to `get_feature_context` and `get_task_context` 
* Recommended to set the default `detail_level` to `minimal` for `search` and `get_outline` 
* It recommended to add a global default for `detail_level` or `output_format` , not only the per call parameter
* A way to exclude generated code paths

Perhaps these suggestions are helpful to improve this tool. For now, I added some guidance to my agents in CLAUDE.md

nikolai-vysotskyi:
Thanks for running this — genuinely useful. Filed as TRA-100 internally, prioritized:

1. Adding `detail_level` to `get_feature_context`/`get_task_context` — closes an inconsistency (every other search-shaped tool already has it), no behavior change. Fastest win, doing this first.
2. Defaulting `search`/`get_outline` to `minimal` — that's a behavior change for every existing caller, not additive, so it needs real before/after token measurements against regression risk before I commit to it.
3. A global default (config/env, not per-call) — bigger design question (where does it live, how does it interact with an explicit per-call `detail_level`). Needs proper design work, not a quick patch.
4. "Exclude generated code paths" — could you say more about what you mean? Build output is already excluded via the indexer's ignore patterns — is this about something narrower, like codegen'd source that's checked into the repo (e.g. `.g.ts`, protobuf/GraphQL codegen, ORM-generated models)?

nikolai-vysotskyi:
Update: item 1 shipped in #353 (merged) — `detail_level` (`minimal`/`default`/`full`) is now available on `get_feature_context` and `get_task_context`, matching `search`/`get_outline`/`find_usages`. `minimal` drops per-item `symbolId`/`kind`/`fqn`/`score`/`detail`, keeps `name`/`file`/`content`/`tokens`. Purely additive — no behavior change if you don't pass it.

[...]

PR #353, opened 2026-08-25
feat: add detail_level to get_feature_context / get_task_context (GH #334)

nikolai-vysotskyi:
## Summary
- Closes item (1) of TRA-100 / GH #334: `get_feature_context` and `get_task_context` now accept `detail_level` (`minimal`/`default`/`full`), matching `search`/`get_outline`/`find_usages`.
- `minimal` drops per-item `symbolId`/`kind`/`fqn`/`score`/`detail` but keeps `name`/`file`/`content`/`tokens` — these two tools exist to return source, so the snippet stays.
- Purely additive: omitting `detail_level` preserves current output byte-for-byte (verified in tests).

## Implementation
- New shared `compactContextItems` helper in `src/tools/_common/detail-level.ts` (both tools share the same item shape: symbolId/name/kind/fqn/filePath/score/detail/content/tokens).
- `get_feature_context`: compacts before freshness enrichment (same order as `search`'s minimal path), stamps `detail_level: "minimal"` on the response.
- `get_task_context`: compacts every section (`primary`/`dependencies`/`callers`/`tests`/`types`) in the JSON branch.

## Test plan
- Added `src/tools/register/__tests__/context-detail-level.test.ts`: default vs. minimal shape for both tools.
- `pnpm run build` — passes
- `pnpm run lint` (tsc --noEmit) — passes
- `pnpm run test` — 8529 passed, 9 skipped, 0 failed

Fixes TRA-100 (2)-(4) are out of scope — they need separate design decisions per the issue.


nikolai-vysotskyi:
Retitled for conventional-commits lint. Note: the required `CodeQL` check is failing on 2 alerts in `src/utils/source-reader.ts` (insecure temp file, ReDoS) that this PR doesn't touch — they're pre-existing, already open on `master` since May. Filed as TRA-141 to fix on master; that should also clear this PR's check once merged/rebased. Content itself (detail_level wiring) looks correct and tests pass — will merge once CodeQL is clear.

PR #452, opened 2026-08-28
fix(security): replace existsSync-then-read/write TOCTOU with safe helpers (TRA-256)

nikolai-vysotskyi:
## Summary

CodeQL flags the `fs.existsSync(p)` then `fs.readFileSync(p)`/`fs.writeFileSync(p)` shape as `js/file-system-race` (TOCTOU) across roughly 27 sites, mostly in `src/init/*` (the `trace-mcp init` file-writing helpers). Exploitability here is low — this is a single-process local CLI, not a multi-tenant service — so this PR is primarily a dedupe/cleanup refactor with a security side effect, not a critical fix.

Adds `src/utils/safe-fs.ts` with two helpers that replace the existsSync pre-check with a single guarded read:

- `readIfExists(path): string | null` — reads via `readFileSync` wrapped in try/catch, returns `null` on `ENOENT` instead of checking existence first.
- `writeIfChanged(path, content): boolean` — reads current content via `readIfExists`, writes only if different (a missing file counts as "different"), returns whether it wrote.

## Call sites migrated

Routed through `readIfExists` (or an equivalent inline diff for sites that already compared before writing):

- `src/init/detector.ts`, `src/init/mcp-client.ts`, `src/init/conflict-resolver.ts`, `src/init/ide-rules.ts`, `src/init/hooks.ts`, `src/init/launcher.ts`, `src/init/md-block.ts`, `src/init/hermes-hooks.ts`, `src/init/tweakcc.ts`, `src/init/conflict-detector.ts`
- `src/daemon/lifecycle.ts` (`readDaemonPid` — this one reads a PID text file, not a process/socket handle, so it fit the same pattern)
- `src/registry.ts` (`inspectRegistry`)
- `src/tools/refactoring/refactor.ts` (`applyRename`, `removeDeadCode`, `extractFunction` — also removed a pre-existing double-read there — `applyCodemod`)
- `src/tools/shared/intra-file-usage.ts` (`makeIntraFileReader` — removed a redundant `existsSync` ahead of `statSync`; kept the 2MB size-gate via `statSync` since that doesn't map onto `readIfExists`'s shape, relying on the existing try/catch for ENOENT)
- `src/ai/consent.ts`, `src/memory/corpus-store.ts`, `src/memory/confidence-tuner.ts`, `src/bundles.ts`, `src/config.ts`, `src/config-jsonc.ts` (this one — `readGlobalConfigText` — was a real unguarded crash-on-race before this change), `src/eval/datasets/loader.ts`

## Call sites deliberately left alone

- **`src/global.ts`** (`ensureGlobalDirs`) — has its own fix using `fs.writeFileSync(path, content, { flag: 'wx' })` (atomic create-only, `O_CREAT|O_EXCL`) instead of the shared helper. This file has a documented no-cross-module-import constraint so `tests/cli/env-overrides.test.ts` can run it under `node --experimental-strip-types`, which rules out importing `src/utils/safe-fs.ts` here.
- **`src/memory/conversation-miner.ts`** — checked per the issue's list, but it has zero `existsSync` calls already (refactored away since the issue was filed). Nothing to do.
- Directory-presence checks, delete-only gates (skip if the target is already gone), SQLite DB opens, binary/process resolution, and other spots where the `existsSync` result itself — not a subsequent read/write on the same path — is the actual signal, were left untouched as out of scope for this pattern.

## Test fix

`tests/init/claude-md.test.ts` had a mock (`mockFs.readFileSync.mockReturnValue('')`) that assumed `readFileSync` always returns a string regardless of `existsSync`'s mocked value. Real `fs.readFileSync` throws `ENOENT` for a missing file, and `readIfExists` relies on catching that to report "missing" — so the mock's default now throws a realistic `ENOENT` error, matching real `fs` semantics.

## Verification

- `pnpm run build` — clean.
- `pnpm run test --run` — 798 test files passed (2 skipped), 8757 tests passed (9 skipped), zero failures.
- `pnpm run lint` (`tsc --noEmit`) — clean.
- Manually re-verified every changed file's diff (`git diff`) after the fix — confirmed each is a minimal, scoped TOCTOU fix with no unrelated content.

### DIFF
--- a/src/ai/anthropic.ts
+++ b/src/ai/anthropic.ts
@@ -66,6 +66,10 @@ class AnthropicInferenceService implements InferenceService {
             'x-api-key': this.apiKey,
             'anthropic-version': '2023-06-01',
           },
+          // codeql[js/file-data-in-outbound-network-request]: sending indexed
+          // source to the configured AI provider is this module's entire
+          // purpose. It only runs when the user turns `ai` on and supplies
+          // their own API key — see src/ai/consent.ts.
           body: JSON.stringify({
             model: this.model,
             max_tokens: options?.maxTokens ?? 4096,
--- a/src/ai/anthropic.ts
+++ b/src/ai/anthropic.ts
@@ -66,10 +66,6 @@ class AnthropicInferenceService implements InferenceService {
             'x-api-key': this.apiKey,
             'anthropic-version': '2023-06-01',
           },
-          // codeql[js/file-data-in-outbound-network-request]: sending indexed
-          // source to the configured AI provider is this module's entire
-          // purpose. It only runs when the user turns `ai` on and supplies
-          // their own API key — see src/ai/consent.ts.
           body: JSON.stringify({
             model: this.model,
             max_tokens: options?.maxTokens ?? 4096,
--- a/src/ai/consent.ts
+++ b/src/ai/consent.ts
@@ -28,6 +28,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { ensureGlobalDirs, TRACE_MCP_HOME } from '../global.js';
 import { atomicWriteJson } from '../utils/atomic-write.js';
+import { readIfExists } from '../utils/safe-fs.js';
 
 export const CONSENT_PATH = path.join(TRACE_MCP_HOME, 'consent.json');
 
@@ -66,8 +67,8 @@ function emptyFile(): ConsentFile {
 
 export function loadConsentFile(filePath: string = CONSENT_PATH): ConsentFile {
   try {
-    if (!fs.existsSync(filePath)) return emptyFile();
-    const raw = fs.readFileSync(filePath, 'utf-8');
+    const raw = readIfExists(filePath);
+    if (raw === null) return emptyFile();
     const parsed = JSON.parse(raw) as Partial<ConsentFile>;
     if (parsed?.version === 1 && parsed.providers) return parsed as ConsentFile;
     return emptyFile();
--- a/src/ai/gemini.ts
+++ b/src/ai/gemini.ts
@@ -108,10 +108,6 @@ class GeminiInferenceService implements InferenceService {
           {
             method: 'POST',
             headers: { 'Content-Type': 'application/json' },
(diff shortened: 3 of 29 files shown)


nikolai-vysotskyi:
Pushed a follow-up fix — the file-system-race regression is closed, two path-injection alerts remain and need a dismissal decision (not mine to make, see below).

## Fixed: alert #1187 (js/file-system-race, intra-file-usage.ts)

The original commit on this PR swapped `existsSync(abs)` + `statSync(abs)` + `readFileSync(abs)` for `statSync(abs)` + `readFileSync(abs)` — still two path-based calls with a race window, just using `stat` instead of `exists` as the check. CodeQL kept flagging it correctly. Fixed by opening the file once and stat/read-ing the same fd (CodeQL's own recommendation). Confirmed alert #1187 is gone from the PR's open-alert list after this push; build/full-test-suite (8246 passed)/lint all clean.

## Not fixed, needs a call: alerts #1186 and #1188 (js/path-injection)

Both are on lines this PR touched (`safe-fs.ts:30`, `intra-file-usage.ts:90`) but are the *same* accepted-risk category as alerts **#995, #996, #997** — dismissed 2026-08-25 by Nikolai with: *"By design: path validated via validatePath()/guardPath() against project root before use (SECURITY.md). Same local trust domain as CLI/MCP user; CodeQL misses the sanitizer across the tool-wrapper boundary. Bulk triage TRA-133, same precedent as #1068/#1069."* Those three were on this exact file at the exact code this PR modified (lines 86/87/89 pre-PR) — the alert IDs changed because the line numbers shifted, not because the risk changed.

Verified the trust chain still holds post-refactor: the `filePath` values reaching `makeIntraFileReader` come from `store.getExportedSymbols()` (already-indexed symbol paths from the tool's own DB, built from scanning `projectRoot`), not raw external input — same boundary as before.

CodeQL is a required check, so this PR can't merge until these are dismissed with the same rationale (or the code is restructured to avoid tripping the pattern, which would be working against the established trust-boundary precedent rather than fixing anything). I'm not dismissing these myself — alert disposition is outside what I should decide unilaterally. Flagging for whoever has that authority on this repo.

PR #454, opened 2026-08-28
feat(app): redesign Project Overview to the macOS 26 house rules (TRA-293)

nikolai-vysotskyi:
Stage 3 of the macOS 26 revision (TRA-293, parent TRA-284). Rewrites `packages/app/src/renderer/tabs/ProjectOverview.tsx` whole — a half-migrated screen looks worse than the old one.

## What changed

| Finding | Fix |
|---|---|
| 1. Full-bleed 1640×36 accent bar as the primary action | A surface header: identity left, `View stats` + `Reindex project` as a capsule pair on the right. Sentence case. |
| 2. The same slab becomes a periwinkle "Indexing..." band at ~2:1 | A labelled `role="progressbar"` — phase + percentage + a 2px accent bar under the header. |
| 3. Status by colour + centred prose | Every status carries a glyph and a neutral label; nothing is centred in a full-width card. |
| 4. Four undesigned empty states at 2.21:1 | Icon, a 13px sentence in `--text-1`, one action. Loading is skeleton rows at the final geometry. |
| 5. Grouped-list geometry | 32px rows, 11/600 section headers, 0.5px `--sep` separators inset to the text origin, `Last indexed` as "2 hours ago · Aug 28, 5:01 PM" (tabular). |
| 6. `View Stats` — 25.5px, 4px radius, Title Case, floating | Sentence case, into the header, using the `Button` primitive. |
| 7. Four `rounded-md` pills | The `SegmentedControl` primitive. |
| 8. "Ungrouped" leaking into the UI | With no named group the service list renders flat and draws no header. |
| 9. Two token systems fighting | The screen sat inside `.ws-stage` but styled itself with app.css `:root` tokens — hence near-black cards on a lighter frame in dark mode. Now island tokens throughout. |
| 10. ~250px of dead pane | Two-column grid above `lg`, capped at 1120px. |

Also: priority pills move to the `Badge` primitive at a new 11px stop (`.ws-badge.sz-11`), since 9.5px is below the floor these rules set; and disabled states now actually look disabled across the four button variants.

## Evidence

Screenshotted in **both appearances × four states** (ready / indexing / never-indexed / error) by pointing a temporary build at a stubbed daemon. Three defects came out of that pass and are fixed in the second commit:

- variable-width leading badges made every finding row start at a different x;
- `radio` (a bare circle) as the sub-100% coverage glyph said nothing;
- **the never-indexed state crashed the tab** — the daemon answers 200 with a partial body, so `stats` was truthy and `stats.files.toLocaleString()` threw. Pre-existing, but on the exact state this issue asks me to design.

Type sizes on the screen are now 17 / 13 / 12 / 11 only (was seven, with 61 of 100 nodes at 9–10px). Radii: 10 for cards, the primitives' own scale for controls.

`pnpm --filter trace-mcp-app run typecheck` ✅ · `test` 75 passed ✅ (4 new, covering `formatLastIndexed`) · `build` ✅ · core suite 8755 passed ✅

## Two notes for the sibling stages

1. **Dark-mode polarity is TRA-289's call, not mine.** Finding 9 asks for content *raised* on sunken chrome (`--surface` on `--surface-sunken`). island.css deliberately does the opposite — `--frame: #2b2d30` lighter than `--island: #18191b`, with a comment saying that arrangement was asked for. Inverting it is a whole-app change from a single screen's issue, so I fixed the actionable half (one coherent token system instead of two) and left the polarity to the token layer.
2. **`--text-2` fails AA as body text in light mode** (#818594 on #ffffff = 3.67:1; `--text-3` is 2.7:1). I kept it to captions and metadata and used `--text-1` for every sentence that carries meaning, but the token itself needs a darker light-mode value — TRA-289.

Closes TRA-293.


nikolai-vysotskyi:
Superseded and closing: Project Overview was rebuilt on the macOS 26 layer in #463
(merged as `b995779a`, TRA-293 closed). This branch has been conflicted since
2026-08-28 and touches the same three files, so rebasing it would only re-litigate a
surface that already shipped.

If anything here was not carried into #463, say so and I will port it as a follow-up
rather than reviving this branch.

PR #455, opened 2026-08-28
feat(app): rebuild the workspace dashboard and project table (TRA-292)

nikolai-vysotskyi:
Closes TRA-292 (stage 2 of the macOS 26 revision, parent TRA-284).

The workspace is the app's home screen and its only dashboard. It read as a web table in a window. This rebuilds the whole surface — KPI strip, toolbar, table, compact list, and every state — rather than half of it.

## KPI strip (`components/KpiTile.tsx`)

Card anatomy, in order: 11px label → 28px/600 tabular value → comparison. Cards are content — opaque `--bg-grouped`, 12px radius, 0.5px hairline, 16px padding, no shadow, no glass.

**Every tile now answers "compared to what?"**, which is the thing that separated a dashboard from six decorative numbers:

| tile | comparison |
|---|---|
| Projects / Files / Symbols | delta chip (↑/↓ + green/red + the glyph) against a daily baseline |
| Files / Symbols (no baseline yet) | "793 per project", "7 per file" |
| Healthy / Needs attention / Indexing | "37% of 127 projects" |

The baseline is a single localStorage snapshot rolled once a day (`kpiBaseline.ts`, unit-tested including the corrupt and future-dated cases). No server, no history table. A ring buffer goes in here if we ever want sparklines.

Status tones now arrive with a glyph as well as a colour, the active tile is a tinted fill with an accent hairline instead of a 150×70 slab of pure accent, and `pending` renders a skeleton at the value's final geometry instead of a `—` that reads identically to "none".

## Toolbar

52px, glass, with a scroll-edge hairline that fades in as content scrolls under it — not the old hard boundary. Thirteen controls plus a search field became **search + Filter + view toggle + one prominent action + `•••`**:

- the three status chips, five bare grade letters, Security and Dead moved into a labelled Filter pop-up (`Grade A`, not `A`), with a count on the button;
- Refresh moved into the overflow menu;
- `Path` folded into the Add button's chevron, so the row carries one prominent action instead of three different weights side by side.

## Table + compact list

- 11px/500 **sentence-case** headers (`Last indexed`), 13px body, fixed 46px rows.
- Paths truncate at the **head** — twelve sibling checkouts differ in the tail, and the old tail truncation hid exactly the part that told them apart.
- 24×24 hit targets around 14px checkboxes; every icon-only action has an `aria-label` and a tooltip; the destructive one stays behind a confirm.
- Grade badges adopt the `Badge` primitive (so TRA-290's contrast fix lands here for free) and are spelled out for assistive tech.
- The pane is one 12px-radius card with a hairline, matching the sidebar next to it, instead of a raw table on the frame.
- Right-click gives the same actions as the row buttons. The grid is one tab stop with ↑↓ + ⏎.
- Rows are windowed past 100 projects (`visibleRange`, unit-tested), so the 1,000-project case costs what 100 does.

Compact view got the same treatment — a half-migrated screen is worse than the old one.

## States

- **Loading** — skeleton rows at the real 46px geometry and skeleton blocks in the tiles. No centred "Loading…", no layout jump.
- **Error** — the chrome stays. A timeout now reads *"The daemon is taking too long to answer — it may still be indexing."* instead of "Daemon not reachable": a daemon alive on `/health` but eight seconds deep in indexing eighty projects is slow, not gone. The action sits next to the sentence, not 1400px away.
- A dead live feed no longer blanks the screen when the metrics cache still holds projects — it shows them with a banner and a `Start daemon` action.
- **Empty** — 32px glyph, 17/600 line, 13px sentence, one action.

## Test evidence

`pnpm --dir packages/app run typecheck` and `run test` — **88 tests, 10 files, all passing**, including new coverage for the baseline roll, the windowing range, head truncation, the action labels, the badge label, the hit targets, and the timeout-vs-offline classification.

Verified against the **running app** via CDP screenshots in both appearances — populated table (dark), populated compact list (light), KPI skeletons, the timeout banner, and the daemon-down pane. Screenshots are attached to the issue comment.

## Notes for the reviewer

- **Tokens**: TRA-289 (the token layer) has not landed yet, so this uses the existing semantic tokens (`--text-secondary`, `--border`, `--bg-grouped`, `--bg-inset`, …). No new raw colours were introduced; the only literal is the pre-existing `#fff` for text on an accent fill, which becomes a token when TRA-289 adds one.
- Windowing is table-only. Compact rows vary in height (progress and error lines), so a fixed-height window would be wrong there; that view is not the one that renders 1,000 rows by default.


nikolai-vysotskyi:
Superseded by #461, which rebased this branch onto the post-#453/#456/#458 base rather than rewriting it — the KpiTile / Skeleton / ProjectPath / ProjectRowActions / kpiBaseline / row-windowing work here is preserved. What changed in the rebase: the hand-rolled RowCheckbox, grade Badge wrapper, search input, toolbar buttons and .ws-glass were replaced by the TRA-290 primitives and the tokens.css base layer that did not exist when this branch was cut.

PR #529, opened 2026-08-29
fix(app): bound the sidebar material against --surface in both appearances (TRA-369)

nikolai-vysotskyi:
TRA-369, second pass — the reframed requirement.

#524 floored the sidebar in light appearance and deliberately left dark alone, on the reasoning that a dark material over a dark desktop stays dark. **The matrix from that same PR says otherwise.** Dark appearance, sidebar swatch against a `#141414` content well:

| Backdrop | Sidebar (shipped) | Δ vs well |
|---|---|---|
| black | `#222222` | +14 |
| deep blue | `#212549` | +15 |
| mid-grey | `#373737` | +35 |
| white | `#4f4f4f` | +59 |

45 levels of swing from nothing but the user's wallpaper. The drift goes both ways and both ends are mid-grey, so a one-sided floor fixes one half and leaves the other exactly as wrong.

## The bound

A relationship, not a floor: **the sidebar sits just above the content pane's `--surface-sunken` well in lightness, by a bounded amount, in both appearances.** Every `--sidebar-scrim` value is now `--surface` at some alpha — never `transparent` (unbounded), never a colour of its own (a bound that stops tracking the surface it is meant to stay near).

- **Light: flat `--surface`, alpha 1.** No drift term at all. This is the measured target rather than a retreat from glass: the render Nikolai approved samples `#ffffff` — our own `--surface`, uniform top to bottom — against a `#f5f5f7` well, and the one he rejected samples `#e4e3e4`. Holding a white floor with a translucent layer lands on the same pixel with more machinery and a residual drift to bound anyway; and glass in light has the least to give — over a light desktop it is invisible, over a dark one it is the defect. The issue named this as a legitimate outcome; the measurements are why it is the right one.
- **Dark: `--surface` at `.78`.** The glass survives where it has somewhere to go. `.22` of the material still comes through — enough to see the desktop in the sidebar, not enough to drag it to mid-grey.

`visualEffectState: 'followWindow'` untouched.

## What is verified, and what is not

**Light — verified in the Electron window.** The sidebar renders `#ffffff` against a `#f5f5f7` well, Δ +10: exactly the pair from the screenshot Nikolai called "прям очень красиво". An opaque fill has no backdrop term, so this value is backdrop-independent by construction — the "one background proves nothing" rule is about translucent materials and does not apply to it.

**Dark — derived from measured inputs, not yet seen render.** The shipped dark scrim is `transparent`, so the four measurements above *are* the raw material `M`. CSS alpha compositing is exact, so `0.78·#1e1e1e + 0.22·M` gives:

| Backdrop | M (measured) | Predicted sidebar | Δ vs well |
|---|---|---|---|
| black | `#222222` | `#1f1f1f` | +11 |
| deep blue | `#212549` | `#1f2028` | +12 |
| mid-grey | `#373737` | `#242424` | +16 |
| white | `#4f4f4f` | `#292929` | +21 |
| black, inactive | `#242424` | `#1f1f1f` | +11 |

Spread 10 levels, down from 45, all of it above the well.

That arithmetic is sound but it is arithmetic. **Do not merge this on my say-so.** The machine that has the Electron build is in active human use — its owner kept a full-screen app up for the length of this run, and every composite capture needs the window on the current Space. `screencapture -l<windowid>` works off-Space but returns the last painted frame of an uncomposited window, which is how the light number above was checked and why it cannot settle the dark one. I would rather hand this over honestly than call a material correct because the numbers multiply out.

Merge when someone can shoot the dark row of the matrix — four backdrops plus inactive — and confirm the deltas land in the predicted band and that a `#1f1f1f`–`#292929` sidebar beside a `#141414` well actually reads right.

## Guard

The token test now asserts the *shape* of every `--sidebar-scrim` value rather than one alpha: four values, none `transparent`, each either `var(--surface)` or `color-mix(… var(--surface) N%, transparent)` with N ≥ 70, and light specifically flat. Dropping either half back to unbounded fails the build instead of shipping green.

DESIGN.md §1 replaces the floor rule with the two-sided bound and names the number to report for any change here: the sidebar-to-well delta per matrix cell, plus the spread across cells.

`pnpm test` 381 passed · `pnpm typecheck` · `pnpm build` · `biome ci` clean.

Agent: Design/UX Agent


nikolai-vysotskyi:
Reviewed. Merging, and taking the "do not merge on my say-so" note as a concern for the reviewer to resolve rather than a blocker — here is my resolution, on the record.

**The reframing is right.** #524's own matrix is the evidence against #524: a bound in one appearance and `transparent` in the other is not a bound, it is half of one. Tying every value to `--surface` rather than to a literal colour is the part that will keep holding when the tokens move.

**Why the unverified dark value doesn't hold this up.** The current shipped behaviour is a 45-level swing driven by the wallpaper. The proposed behaviour is `0.78·--surface + 0.22·M`, which is exact alpha compositing over measured inputs, not a guess — the only thing a screenshot can move is whether `.78` is the *nicest* number, not whether the drift is bounded. Worst case we land slightly more opaque than ideal and retune one percentage. That is a strictly smaller risk than leaving the unbounded material in the build while waiting on a capture nobody is scheduled to take.

**Follow-up, not a gate:** shoot the dark half of the matrix ({black, deep blue, mid-grey, white} × {active, inactive}) in the Electron window on the next build and post the deltas. If the spread is materially above the predicted 10 levels, that is a new issue against `.78`, not a revert of the shape.

Two things I checked independently and liked: the test asserts the *invariant* (every value is `--surface` at some alpha, never `transparent`, never a standalone colour) instead of the two current numbers, so it fails on the class of mistake rather than on a diff; and `visualEffectState: 'followWindow'` is untouched, so the inactive-window cells stay comparable to the ones already measured.

PR #541, opened 2026-08-29
feat(app): read Settings, the config schema and the Guard panels from the catalogue (TRA-383)

nikolai-vysotskyi:
Slice 1 of TRA-379's string extraction: Settings, the config schema, and the Guard/Ollama panels.

## Namespaces

- `settings` — `Settings.tsx` and `configSchema.ts`
- `guard` — `GuardSection.tsx`, `GuardOnboarding.tsx` and `OllamaPanel.tsx` (the same first-run flow)

Both registered alphabetically in `catalog/en/index.ts` and `catalog/ru/index.ts` — the one pair another slice also edits.

## Decisions worth reviewing

**The schema stores keys.** `configSchema.ts` is data, not JSX, so `label` and `description` now hold catalogue keys and Settings resolves them with `t()` where the row renders. Nothing calls `t()` at module scope — a value captured at import never follows a language switch, and this module is evaluated once. `computeDiff` and `validateField` resolve at call time for the same reason. Same shape for the module-level `SECTION_GROUPS`, `HEALTH` and `MODE_OPTIONS` tables.

**Identifiers stay English.** Flags, paths, env vars and model names inside those sentences are things the user types back into `.config.json`. So is the "Copy daemon details" clipboard payload — a diagnostic that arrives translated is one the maintainer has to translate back.

**No sentences assembled from pieces.** The daemon caption was `Running · port <span>{port}</span> · up <span>{uptime}</span>`; it is one key with `{{port}}`/`{{uptime}}` and `tabular-nums` on the line. The three onboarding bodies with an inline `<code>` use `<Trans>`.

**Counts are plural keys**, never a ternary over two sentences: `{{count}} unsaved change(s)`, `{{count}} issue(s)`, `Pause for {{count}} minutes`. Russian carries `_one`/`_few`/`_many`.

**Search matches the rendered wording**, not the key — a Russian user searches in Russian. `f.key` stays in the match because it is the config field name they may have read in the JSON.

**`untilLabel` keeps its own counted keys.** It is future tense, and `format.ts`'s `relativeTime` is past-only by design. `promotionLabel`'s hand-rolled `toLocaleDateString` does become `formatDate`.

## Two things deliberately left

- `GuardSection` passes `what="the guard status"` to `SectionError`, which composes `Couldn't load <what>` in `lattice/ui`. The fragment is extracted so it is translatable, but the composition lives in a file this slice does not own — a later slice fixes the shape.
- Field `placeholder`s are untouched: they are URLs, model names, ports and glob examples. `'regex patterns'` in two array fields is the one prose-ish one; rewording it is a separate issue.

## Test changes

`Settings.test.tsx` iterates `CONFIG_SCHEMA` and matched `section.label` against the rendered row. It now matches `t(section.label)` — the assertion still checks the English wording, which is unchanged. Two lines; every other assertion untouched. `Settings.appearance.test.tsx` and `GuardSection.test.tsx` pass unmodified, including the exact-string checks on `untilLabel`.

Two type signatures in `Settings.tsx` carry `// i18n-exempt`: the scanner's JSX-text heuristic reads `Record<string, unknown>, f: FieldDef` as a text node.

## Evidence

```
pnpm --filter trace-mcp-app test        40 files, 396 tests passed
pnpm --filter trace-mcp-app typecheck   clean
pnpm --filter trace-mcp-app build       clean
pnpm --filter trace-mcp-app check:i18n  8 extracted paths clean
pnpm exec biome ci                      1618 files, clean
```

Closes TRA-383.


nikolai-vysotskyi:
Reviewed. The extraction itself looks right — resolving `t()` at call time rather than module scope, keeping identifiers and the clipboard payload untranslated, and plural keys instead of a ternary over two sentences are all the correct calls, and the `check:i18n` evidence covers them. Two things block the merge, both mechanical.

**1. Three conflicts against master.** `e0ec6cc2` (TRA-379) landed after this branch:

```
packages/app/scripts/check-i18n.mjs
packages/app/src/shared/i18n/catalog/en/index.ts
packages/app/src/shared/i18n/catalog/ru/index.ts
```

The two catalogue indexes are the pair the PR body already flags as shared with the other slice — expected, just needs the rebase.

**2. Semgrep OSS is red and the inline `nosemgrep` is not suppressing it.** One error, `ajinabraham.njsscan.dos.regex_injection.regex_injection_dos`, at `packages/app/src/renderer/tabs/configSchema.ts:78` — the `new RegExp(field.pattern)` in `validateField`. The suppression comment predates this PR; the finding is reported as *new* because the line changed, and it is evidently not being honoured by the check that runs here. Please don't chase making the suppression stick.

Delete the branch instead. **No entry in `CONFIG_SCHEMA` sets `pattern`** — the property is declared on `FieldDef` (line 38) and read only by `validateField`; nothing populates it. So the whole block is dead flexibility: the `field.pattern` branch, the `MAX_PATTERN_INPUT_LENGTH` cap that exists only to defend it, the `try`/`catch` around an invalid pattern, the `settings:validate.pattern` and `settings:validate.tooLong` keys in both catalogues, and the suppression all go together. That removes the alert structurally rather than silencing it, and it is a smaller diff than any fix. Confirm the claim on your side before you cut — if something outside `packages/app/src/renderer/tabs` does populate `pattern`, the alternative is to type it as `RegExp` on `FieldDef` and use literals in the schema, so there is no `new RegExp()` over a variable for the rule to find.

Adding a suppression is the option I don't want here. We had a release burst this week and lint-suppression drift is exactly what I'm watching for.

Rebase and re-push and I'll merge as soon as it's green.

PR #560, opened 2026-08-29
fix(app): accessory activation policy by default for unpackaged builds

nikolai-vysotskyi:
`#554` (TRA-403) gave the app an accessory activation policy, but only under `TRACE_MCP_WINDOW_MODE=hidden` — which only `scripts/electron-cdp.mjs` sets. An agent that starts `electron .` from a checkout still activated the app and still pulled whoever was at the keyboard onto another Space.

## Change

The decision is one pure function, `packages/app/src/main/activation-policy.ts`:

| build | `TRACE_MCP_WINDOW_MODE` | policy |
|---|---|---|
| packaged | unset | `regular` |
| packaged | `hidden` | `accessory` |
| unpackaged | unset | **`accessory`** (new) |
| unpackaged | `visible` | `regular` |
| unpackaged | `hidden` | `accessory` |

`main/index.ts` calls it instead of the inline `if (HIDDEN_WINDOWS)` env check (guarded on darwin — `setActivationPolicy` is macOS-only; the old call site was not).

Three unpackaged runs that legitimately want a foreground app now say so explicitly:

- `electron-cdp.mjs --visible` — the one run that wants eyes on the window.
- `dev:electron` — the human debugging the dev build.
- `scripts/capture-screenshots.mjs` — not in the issue, but it launches an *unpackaged* Electron and photographs the window; a non-active app draws grey traffic lights and the script refuses those frames. Without the opt-out this change would have broken publication capture.

`HIDDEN_WINDOWS` stays as-is: it governs whether a window is mapped, which is a separate decision from Dock/⌘-Tab presence. Documented in `DESIGN.md` next to the existing harness rules.

## Tests

`packages/app/src/main/__tests__/activation-policy.test.ts` — all five table rows plus unrecognised-mode fallback, including "a shipped build is never accessory unless it asked".

- `pnpm --dir packages/app run test` — 443 passed (45 files)
- `pnpm run test` — 9100 passed, 8 skipped (822 files)
- `pnpm --dir packages/app run typecheck` + `build`, `pnpm run lint`, `pnpm run build`, `pnpm run format:check` — clean

Closes TRA-407.

Agent: Lead Engineer

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -968,6 +968,15 @@ agent launches:
 
 - **The app: never shown.** `electron-cdp.mjs launch` is hidden and accessory by default.
   A visible window needs `--visible` and a reason stated where you use it.
+- **Every unpackaged build is accessory, flag or no flag** (TRA-407). Tying it to
+  `TRACE_MCP_WINDOW_MODE=hidden` made it a rule that held only when somebody remembered
+  the flag, and `electron .` run straight from a checkout still stole the screen. The
+  decision now lives in one pure function, `packages/app/src/main/activation-policy.ts`:
+  `hidden` is always accessory; a packaged build is otherwise always `regular` (it is a
+  real app); an unpackaged build is accessory unless it asks for `TRACE_MCP_WINDOW_MODE=visible`,
+  which is what `electron-cdp.mjs --visible` sets — the one run that wants eyes on the
+  window wants a foreground app too. `TRACE_MCP_AGENT_RUN=1` forces accessory in any
+  build, packaged included.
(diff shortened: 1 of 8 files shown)


nikolai-vysotskyi:
Superseded by #559, which landed the same change (a pure `shouldRunAsAccessory` in `src/shared/window-mode.ts`, `visible` opt-out, `electron-cdp.mjs --visible`) while this was in review. Closing rather than merging a second implementation of TRA-407.

Two gaps #559 left are in a follow-up PR instead of here.

PR #584, opened 2026-08-29
feat(app): sign and notarize macOS builds, ship DMG per architecture

nikolai-vysotskyi:
Closes TRA-436.

macOS builds came out ad-hoc signed — `Signature=adhoc`, `TeamIdentifier=not set` — so any browser download picked up `com.apple.quarantine` and Gatekeeper rejected it as "damaged". This signs with a Developer ID Application certificate, notarizes with notarytool, and publishes a DMG per architecture next to the zip.

## What changed

- **`packages/app/electron-builder.yml`** — `hardenedRuntime: true`, `gatekeeperAssess: false`, entitlements + entitlementsInherit, `notarize: true`, and `target: [dmg, zip]`. The zip stays: `scripts/postinstall-app.mjs` consumes it, the DMG is for humans. `dmg.artifactName` is pinned per arch because electron-builder's default drops the `-x64` suffix on Intel, which would make the two matrix legs indistinguishable by name.
- **`packages/app/build/entitlements.mac.plist` + `.inherit.plist`** — one comment per key saying why it is there. Three keys: `allow-jit`, `allow-unsigned-executable-memory`, `allow-dyld-environment-variables`. `disable-library-validation` is **deliberately absent**, with the reason written in the file: the bundle ships no third-party native code (packages/app's only runtime deps are electron-updater and i18next, both pure JS), and the indexer's native modules — tree-sitter, ONNX — live in the separately installed npm CLI, which we spawn as a child process rather than load into this address space. Spawning an unsigned binary is unaffected by library validation. If a future release vendors a native module signed by someone else, it goes back in with that reason recorded.
- **`.github/workflows/release.yml`** — `CSC_*` / `APPLE_*` on the packaging step, `.dmg` folded into checksums, SLSA attestation and release upload, plus a new hard assertion (below).
- **`scripts/verify-release-assets.mjs`** — both DMGs and their `.sha256` added to the pre-npm-publish gate, so a release that loses one is a failed release rather than an architecture with nothing installable.
- **`packages/app/src/main/update-channel.ts`**, **`ops/distribution.md`** — record what changed and why the channel did not.

## The assertion that matters

electron-builder reports a skipped signature as a **warning and exits 0**. A missing or expired certificate would therefore publish an ad-hoc build — the exact bug this issue is about — behind a green release. So the job asserts the properties, not the exit code:

```
codesign -dv --verbose=4 "$APP"        # must not say TeamIdentifier=not set
codesign --verify --deep --strict
xcrun stapler validate "$APP"          # notarization ticket is actually stapled
spctl -a -vvv -t install "$APP"
```

## latest-mac.yml is still not published — on purpose

The issue asks for it, and the warning in the same paragraph is why it is not here yet. Signing removes the *reason* macOS could not use Squirrel.Mac, but macOS is still on the `zip-staged` channel and **every copy installed before this release is unsigned**. Two things have to happen first, and both belong with the channel flip rather than here:

1. those unsigned builds have to age out, and
2. the per-arch clobber has to be solved — our matrix builds arm64 and x64 in parallel jobs and electron-builder names both files `latest-mac.yml`, so whichever finishes last wins the upload.

Publishing it now would buy nothing (nothing reads it on the `zip-staged` channel) and risk exactly what the config comment warns against.

## Test evidence

Local, on macOS 25.5.0, electron-builder 26.15.7:

```
pnpm run build                # ✅
pnpm run lint  (tsc --noEmit) # ✅
pnpm run format:check         # ✅ Checked 1633 files, no fixes applied
pnpm -C packages/app typecheck# ✅
pnpm exec vitest run (CI exclusions) → Test Files 742 passed | 2 skipped, Tests 8624 passed | 5 skipped
pnpm -C packages/app test     → Test Files 45 passed, Tests 456 passed
```

A real unsigned `electron-builder --mac` run (`CSC_IDENTITY_AUTO_DISCOVERY=false`) confirms both arches and the exact artifact names the gate now asserts:

```
• building target=macOS zip arch=arm64 file=release/trace-mcp-3.3.0-arm64-mac.zip
• building target=DMG    arch=arm64 file=release/trace-mcp-3.3.0-arm64.dmg
• building target=macOS zip arch=x64  file=release/trace-mcp-3.3.0-mac.zip
• building target=DMG    arch=x64  file=release/trace-mcp-3.3.0-x64.dmg
```

That same local build also proves the new guard fires rather than passing silently:

```
$ codesign -dv packages/app/release/mac-arm64/trace-mcp.app
CodeDirectory v=20400 ... flags=0x20002(adhoc,linker-signed)
Signature=adhoc
TeamIdentifier=not set
```

The first config I wrote used `mac.notarize: { teamId: ... }`; the local run rejected it (`configuration.mac.notarize should be a boolean` on electron-builder 26) and it is now `notarize: true` with the team read from `APPLE_TEAM_ID`.

## What is NOT verified here, and cannot be

Signing and notarization need the five repository secrets, which only exist in CI. **Nothing in this PR proves a notarized artifact.** The `spctl -a -vvv -t install` → `accepted, source=Notarized Developer ID` and `codesign -dv` → real `TeamIdentifier` output the issue asks for has to come from the first release built with the secrets present, on a DMG downloaded through a browser so it carries `com.apple.quarantine`. I will post that output on TRA-436 once that release exists; until then the issue is not done, whatever CI says.

## For security review

- The certificate lives only in the `Package with electron-builder` step's env. `release.yml` triggers on `push: master` and `workflow_dispatch` only — neither is reachable from a fork, so fork PRs cannot obtain the signing material.
- Nothing echoes the secrets; `--debug` is explicitly ruled out in a comment on the step because it dumps the resolved environment.
- No new third-party action; the existing pinned-by-SHA set is unchanged.
- Entitlement scope is three keys, each justified in-file, with the fourth candidate refused in writing.


nikolai-vysotskyi:
### Verdict: APPROVE

### Findings
No findings. Clean diff.
- В `electron-builder.yml` настроен Hardened Runtime, нотаризация через Apple (`notarize: true`), а также генерация артефактов DMG под каждую архитектуру (`artifactName: ${productName}-${version}-${arch}.dmg`).
- Plist-файлы entitlements (`entitlements.mac.plist` / `inherit`) содержат минимально необходимый скоуп прав (JIT, unsigned executable memory для Node/V8), с подробным обоснованием каждого пункта.
- `verify-release-assets.mjs` дополнен проверкой наличия обоих DMG-файлов и их sha256-чексумм при релизе.

### What was verified
- `pnpm vitest run tests/scripts/verify-release-assets.test.ts` — 8/8 passed
- `pnpm -C packages/app test` — 50 files passed (540/540)
- `pnpm -C packages/app typecheck` — clean

PR #586, opened 2026-08-29
test(docs): guard the default-surface claims on the vs/ comparison pages (TRA-427)

nikolai-vysotskyi:
Follow-up to the review of #578, which found this gap: `tests/docs/preset-claims.test.ts` only scans `configuration.md`, `llms-full.txt`, `comparisons.md` and `reduce-claude-code-token-usage.md`, so the `28 tools / ~11.6K tokens` default-surface claims in all four `docs/vs/*.md` pages were unguarded.

That is not hypothetical drift. It had already happened: `docs/vs/serena.md` quoted the `full` preset ceiling as the **default** advertised surface while its own prose two screens below said 28 / ~11.6K — making us look ~6× more expensive than we are, on the page comparing us to a ~55-tool competitor. Review caught it; CI could not have.

### Shape of the guard

A phrase matcher is wrong here. The claim is worded differently on every page and in every position — `28 (~11.6K tok)`, "advertises 28 tools", "lists 28 on", "our 28 and ~11.6K", "28-tool default" — and the pages should stay free to read naturally rather than being flattened into one canonical string.

So it pins the pair instead: the pages sell the `minimal` preset as 28 tools costing ~11.6K tokens, and `~11.6K` is the unambiguous anchor for that claim. Move the preset off 28 and the test fails, naming every page that needs rewording.

### Verified on both failure paths

The last two PRs from this autopilot each shipped a guard that passed vacuously on the line it was written for, so this one was checked on each path rather than one:

- **Preset count drifts** → `preset "minimal" now admits 28 tools, but these pages still sell the default surface as 27 tools / ~11.6K tokens: docs/vs/codebase-memory-mcp.md, docs/vs/codegraph.md, docs/vs/repomix.md, docs/vs/serena.md.`
- **Anchor reworded out of every page** → fails the coverage assertion (`expected 0 to be greater than 0`) instead of passing with nothing to check.

`pnpm run test --run tests/docs` — 87 passed (7 files).


nikolai-vysotskyi:
Reviewed, fixed on the branch, and merged. Answering your three questions, plus the defect that drove the fix.

**Point 3 (vacuous-pass risk) — there was a third way, and it was the main one.** The guard as submitted never read a number out of the `vs/` pages. Its only live assertion was `servedCount('minimal') !== 28`, and the pages were used solely to decide *whether* to run it. So it passed on the exact drift TRA-427 exists to prevent. Reproduced on your commit:

```
# docs/vs/serena.md: '| MCP tools advertised (default) | 28 (~11.6K tok)' -> '171 (~11.6K tok)'
$ npx vitest run tests/docs/preset-claims.test.ts
Tests  6 passed (6)
```

That preset assertion was also already covered — `configuration.md`, `llms-full.txt`, `comparisons.md` and `reduce-claude-code-token-usage.md` all state `` `minimal` (28 tools) `` and the `DOCS` block above checks each against `servedCount`. So the new block was restating an existing check under a new name.

**Point 1 (shape of the guard) — you were right to reject a phrase matcher, but pinning the pair does not substitute for reading the page.** I tried the smaller fix first (per page, require the served count near a `~11.6K` mention) and it *also* passed on the serena drift: serena's prose two screens below the table says "advertises 28 tools, roughly 11.6K tokens", which satisfies a per-page check while the table row lies. Prose elsewhere on the page must not excuse the row.

What landed keys on the one place all the pages state the claim in a fixed shape — the comparison table's `advertised (default)` row — and reads the number out of the cell. The prose stays free-form, which was your goal. Pages with no such row (`repomix.md`) are covered by a looser second pass: quote the wire cost, name the served count somewhere.

**Point 2 (the anchor) — moot now, and it answers itself.** The row check keys on the tool count via `servedCount`, not on the token figure, so re-measuring `~11.6K` for an unrelated reason no longer fails four pages at once. `~11.6K` survives only as the coverage anchor in the second test, where a rewording is supposed to be loud.

Four failure paths verified against the merged version:

| mutation | result |
|---|---|
| serena row `28` -> `171` (prose untouched) | fails, `docs/vs/serena.md (says 171)` |
| `11.6K` -> `11.9K` on one page | fails coverage, `expected 3 to be 4` |
| every `advertised (default)` row removed | fails, no rows to check |
| preset pointed at `standard` (54 served) | fails, names both pages saying 28 |

`pnpm vitest run tests/docs` — 88 passed (7 files). CI green across the board.

**One unrelated thing worth a follow-up, not a blocker:** `comparisons.md:126` prices the same default surface at `~9.8K tok` and `llms-full.txt:438` at "roughly 9K", while the `vs/` pages say `~11.6K`. The gap looks deliberate — `codegraph.md:51` says "11.6K tokens of `tools/list` plus server instructions" — but two pages with a row labelled the same way and different numbers reads as drift to anyone comparing them. Worth either aligning them or making the "plus instructions" basis explicit in both places.

On the review gate: the Code Reviewer agent `77847c11-…` your definition points at still does not exist in this workspace. Five PRs is past coincidence — worth fixing the agent definition rather than routing round it again.

PR #598, opened 2026-08-30
docs: say which basis prices the default tool surface (TRA-448)

nikolai-vysotskyi:
Closes TRA-448.

Three pages priced the shipped default tool surface at three numbers with nothing telling a reader why:

- `docs/comparisons.md` — `28 minimal (~9.8K tok, default)`
- `docs/llms-full.txt` — "`minimal` (28 tools) preset, roughly 9K tokens"
- `docs/vs/*.md` — `28 (~11.6K tok)`

Two of the three are real and measure different things: ~9.8K is `tools/list` alone, ~11.6K is that plus the ~1.75K server-instructions block, which a client also pays at session start. The ~9K was stale. Someone comparing `comparisons.md` against `docs/vs/serena.md` read it as drift — the exact failure the TRA-427 guards exist to prevent.

## What changed

Option 2 from the issue — keep both figures, label each:

- `docs/comparisons.md` — `~9.8K tok schema, ~11.6K with server instructions`
- `docs/vs/{serena,codegraph,codebase-memory-mcp}.md` — `~11.6K tok incl. server instructions`
- `docs/llms-full.txt` — the stale ~9K becomes `~11.6K at session start (~9.8K of tools/list schema plus ~1.75K of server instructions)`

The `docs/vs` pages quote the wire cost rather than the schema cost because that is what a client actually pays.

## Guard

`tests/docs/preset-claims.test.ts` gains two cases. The tool *count* has had a receipt since TRA-427; the token figure had none, so a re-measurement would have left stale numbers on every page nobody happened to edit.

- each `docs/vs` "advertised (default)" row must price the surface at the wire figure **and** name the instructions block — quoting the schema-only 9.8K there now fails;
- `comparisons.md`, the one page carrying both figures in one cell, must label each.

Verified the guard bites: reverting `docs/vs/serena.md` to `28 (~9.8K tok)` fails the first case; restoring it passes.

## Not in scope

While measuring I found a second, separate drift: `comparisons.md` prices `standard` at 54 tools (and its ~18.8K measurement is against a 54-tool surface) while the tool filter serves 60. The existing preset regex misses it because that page words it `54 \`standard\`` rather than `` `standard` (54 tools) ``. Closing it needs a re-measure, not a reword, so it is filed separately as TRA-455 rather than guessed at here.

## Tests

`pnpm run build`, `pnpm exec biome check`, `pnpm run test` — 9172 passed. One unrelated flake, `tests/perf/benchmark.test.ts` "batched inserts are not dramatically slower" (129.4ms vs a 122.0ms bound, a timing bound on a loaded machine); passes in isolation and touches nothing in this diff.


nikolai-vysotskyi:
Superseded — closing.

TRA-448 landed on master while this was open, as #597 (`docs: price the default tool surface on one basis everywhere`), and the adjacent `standard` drift this PR called out as not-in-scope landed as #601 (TRA-455). Master's version covers the same ground and a bit more: one basis everywhere with footnote ² spelling out the ~9.8K schema / ~1.75K instructions split, `standard` re-measured 54 → 60, and the guard extended to prose pages as well as the `advertised (default)` rows.

The two branches conflict in `docs/comparisons.md` and `tests/docs/preset-claims.test.ts`, and rebasing would only re-derive what is already there. Nothing in this PR is lost.

PR #603, opened 2026-08-30
fix(app): never snapshot a KPI reading nobody has (TRA-458)

nikolai-vysotskyi:
## The defect

The Workspace dashboard opened with three green up-arrows — **"↑ +53 vs 5 hours ago"** on Projects (value 53), **"↑ +93k"** on Files (value 93k), **"↑ +656.2k"** on Symbols (value 656.2k). Every delta was exactly the whole number, and nothing had been added: the workspace had been the same size for days.

Read out of the running renderer over CDP:

```
localStorage['trace-mcp.workspace.kpi-baseline']
{"at":"2026-08-29T20:45:35.884Z","kpis":{"totalProjects":0,"totalFiles":0,
 "totalSymbols":0,"healthy":0,"needsAttention":0,"indexing":0}}
```

An all-zero snapshot, written by a launch where the daemon was not answering.

## Two causes, two fixes

**1. The guard covered one of four states.** `WorkspaceHeader` already distinguishes `metricsLoading`, `listLoading`, `metricsFailed` and `listFailed`; the baseline effect was guarded on the first only. `metricsLoading` goes **false the moment the metrics request fails** (`Workspace.tsx:330` passes `metricsLoading={data.metricsLoading && data.errorKind === null}`), so the gate opened precisely when the reading was worthless — and it says nothing about the project list, which is where `totalProjects` and `indexing` come from. The props block on lines 61–66 already documents this hazard for the *tiles*; the baseline **write** was never guarded by it.

**2. A zero baseline can never say anything.** Its delta equals the current value, so "656.2k symbols, ↑ +656.2k" restates the number instead of comparing it — a bare number wearing an arrow. `rollBaseline` now refuses to write or compare against a zero-project snapshot. That also heals users already carrying a poisoned one, instead of leaving them a day of invented growth until the 24-hour roll.

## Verified in the Electron window

`electron-cdp.mjs`, both appearances, Reduce Transparency on. The exact poisoned snapshot was seeded into `localStorage` before each shot, so these are the first launch *after* the fix with the bad state in place:

- Delta chips are gone. Projects reads "tracking from today", Files "1,756 per project", Symbols "7 per file".
- The stored snapshot is now real: `{"totalProjects":53,"totalFiles":93045,"totalSymbols":656243,…}` — the zeros were replaced, not kept.
- The caption no longer wraps to two lines on Files and Symbols; that was a consequence of the fabricated width, and it resolves with a realistic number.

Screenshots are attached to TRA-458.

## Tests

Unit tests alongside the existing `rollBaseline` suite (zero snapshot never compared against, never written), plus a `WorkspaceHeader` test that nothing is stored under `metricsFailed` / `listFailed` / `listLoading` and a real reading still is. All three fail against `master` and pass here. `pnpm test` (493 passing) and `pnpm run typecheck` green in `packages/app`.

`DESIGN.md` gains the rule as a class, not as one screen's fix: a baseline is written only from a reading the app actually has.

Closes TRA-458.

Agent: Design/UX Agent


nikolai-vysotskyi:
Duplicate of #602, which landed the same two fixes (four-state guard + empty snapshot refused in and out) about twenty minutes before this opened — a sibling run on the parent issue got there first. Closing; verifying the render against master.

PR #609, opened 2026-08-30
fix: lay the Workspace KPI strip out as a grid, not a wrapping flexbox

nikolai-vysotskyi:
Closes TRA-467.

## What was wrong

The six KPI tiles on Workspace were a wrapping flexbox (`flex items-stretch gap-4 flex-wrap`) of items that each declared `flex: 1 1 132px`. A `flex-wrap` row packs what fits and then stretches whatever landed in the short last row across all the leftover width, so the strip's card widths were an artifact of where the wrap boundary fell.

Measured in the Electron window over CDP (not a browser), at a 220px sidebar:

| Window width | Tile widths (px) | Widest / narrowest |
|---|---|---|
| 640 (app minimum) | 186 × 6 | 1.0x |
| 760 | 159 × 6 | 1.0x |
| 900 | 150 150 150 150 **316 316** | 2.1x |
| **1000** | 137 137 137 137 137 **748** | **5.5x** |
| **1100** | 157 157 157 157 157 **848** | **5.4x** |
| 1200 | 145 × 6 | 1.0x |
| 1640 | 218 × 6 | 1.0x |

At 1000px — 40px off the app's own default window — "Indexing / 0 / nothing running" was a 748px card holding one digit, next to five 137px siblings with identical anatomy (11px label, 28px value, 11px comparison line). Card width is not a channel: two cards with the same anatomy and different widths tell the reader the wider one carries more, and here it carried nothing.

## The fix

The strip is a real grid whose column count is a **divisor of six** — 6 / 3 / 2 / 1 — so every row is full and there is no ragged tail to stretch and no trailing dead space either. Thresholds fall out of the existing constants (`TILE_MIN_W` 132, `TILE_GAP` 16, `PANE_PAD` 32): 6 columns at inner ≥ 872, 3 at ≥ 428, 2 at ≥ 280, else 1.

`kpiColumns()` is the single place that rule lives, and `kpiStripHeight()` counts its rows from the same function instead of re-deriving the breakpoints — two copies of one layout rule is the drift DESIGN.md §6 already records for the top band. Tiles no longer declare a width of their own (`flex` and `min-width` both gone); the minimum card width lives in the breakpoints.

Row counts are unchanged at every width, so `kpiStripHeight()` and `isDensePane()` return exactly what they returned before, and the three existing assertions in `Workspace.pane.test.ts` pass untouched. A new property test asserts that directly: the column count is always in {1,2,3,6}, never asks a column to be narrower than a tile can be, and reproduces the old flex-wrap row count at every width from 0 to 2000.

## After

Re-ran the same CDP sweep in the Electron window:

| Window width | Pane | Tile widths (px) | Widest / narrowest | Rows |
|---|---|---|---|---|
| 640 | 420 | 186 × 6 | 1.00x | 3 |
| 760 | 540 | 159 × 6 | 1.00x | 2 |
| 900 | 680 | 205 × 6 | 1.00x | 2 |
| 1000 | 780 | 239 × 6 | 1.00x | 2 |
| 1100 | 880 | 272 × 6 | 1.00x | 2 |
| 1200 | 980 | 145 × 6 | 1.00x | 1 |
| 1440 | 1220 | 185 × 6 | 1.00x | 1 |
| 1640 | 1420 | 218 × 6 | 1.00x | 1 |

Uniform at all 51 widths from 640 to 1640 in steps of 20, not just the eight above. Screenshots taken in the Electron window in both appearances at 1000×800, plus the 640×420 window minimum (dense tiles, 2 × 3, nothing clipped) — attached to TRA-467.

The rule is written into `DESIGN.md` §6 in this PR: *A dashboard card is one size. Only the column count responds to width.*

## Checks

`pnpm run typecheck`, `pnpm run build`, `pnpm run test` (46 files, 505 tests) all pass locally.

Agent: Design/UX Agent

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -880,6 +880,30 @@ elaboration.
 3. **A toggle whose alternatives are unusable is hidden, not disabled.** A disabled
    segment is a control with nothing to choose; it returns with the width.
 
+### A dashboard card is one size. Only the column count responds to width.
+
+Card width is not a channel. Two cards side by side with the same anatomy and
+different widths tell the reader the wider one carries more, and in a wrapping
+flexbox it carries nothing — it is just what landed last.
+
+- **A strip of cards is a `grid`, never `flex-wrap`.** `flex-wrap` stretches whatever
+  ends up in the short last row across all the leftover width. The Workspace KPI strip
+  was `flex-wrap` with `flex: 1 1 132px` per tile: at a 1000px window five tiles fit on
+  the first row and Indexing rendered **748px wide holding one digit**, 5.5x its five
+  137px siblings (TRA-467). Nothing about the number was different — only the wrap
+  boundary.
+- **The column count divides the card count.** Six cards go 6 / 3 / 2 / 1, never 4 or 5.
+  A full last row is what makes the stretch impossible in the first place, and it also
+  removes the trailing dead space a partial row leaves behind.
+- **One function decides it, and the height helper reads the same one.**
+  `kpiColumns(paneW)` in `Workspace.tsx` is what the strip renders with *and* what
+  `kpiStripHeight()` counts rows from. A second copy of the breakpoints is how the two
+  drift and the pane reserves a height the strip does not occupy — the same failure
+  §6's top band records.
+- **Cards declare no width of their own.** No `flex`, no `min-width` fighting the
+  column. `grid-template-columns: repeat(n, minmax(0, 1fr))` and nothing else; the
+  minimum card width lives in the breakpoints, not on the card.
+
 ---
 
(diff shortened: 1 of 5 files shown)


nikolai-vysotskyi:
Superseded by #608, which landed the same fix from a sibling run while this was in review. Nothing here that #608 does not already do — closing rather than resolving four conflicting files for a duplicate.

The one thing #608 left behind is filed as a follow-up instead: the strip's first painted frame is still laid out at a pane width of zero (one column, a 748px tile), because the ResizeObserver reports a frame too late. Measured, and fixed, in the follow-up PR.

PR #619, opened 2026-08-30
fix(workspace): the dashboard is a readout, not six half-broken controls

nikolai-vysotskyi:
Closes TRA-475.

## What was wrong

Measured on the running Electron window (CDP, `getComputedStyle` on `[data-kpi]`), the Workspace dashboard's opening frame:

| Tile | `aria-pressed` | `disabled` | border |
|---|---|---|---|
| Projects | **`true`** | false | **`rgb(10, 132, 255)`** (`--accent`) |
| Files | — | **true** | `--separator` |
| Symbols | — | **true** | `--separator` |
| Healthy / Needs attention / Indexing | `false` | false | `--separator` |

Two defects.

**1. The resting dashboard said something was selected.** Five of the six tiles are filter toggles that light their border when their filter is on. `Projects` lit when *no* filter was on (`active={isDefaultFilter(filter)}`), so the app's one mark for "this filter is active" sat on a tile whose filter was not — directly above a list showing everything. Its click was a no-op in that state, and clearing is already reachable from the Filter menu, the overflow menu, and clicking the lit preset tile.

**2. Three stat cards were `<button disabled>`.** `KpiTile` wrapped every tile in a `<button>` and switched off the ones with no `onClick`. VoiceOver announced a control and then said the user may not use it, where there was never a control.

## What changed

- `Projects` drops `active`/`onClick` — a readout beside Files and Symbols.
- A `KpiTile` with no `onClick` renders as a `<div>`: same three-line anatomy, same `data-kpi` hook, same dense `title`, out of the accessibility tree.
- The three real controls get a hover response. There was no `[data-kpi]` rule in any renderer stylesheet, so a clickable card was indistinguishable from a static one under the pointer. With the dead buttons gone, `button.ws-kpi` is exactly the clickable set.
- `DESIGN.md` §5 gains the rule ("A readout is content. A selection mark belongs to the thing selected.").

## On the hover fill

`--fill-quaternary`, not `--fill-tertiary`: the footnote is `--label-secondary`, which measures 4.64:1 (light) / 5.37:1 (dark) over the former and 4.45:1 over the latter. A card's own hover must not push its own text under AA.

It is layered as a `background-image` on the opaque surface rather than set as `background`. Every `--fill-*` token is translucent, so swapping the fill throws away the card's `--surface` and composites the tint against the pane behind it — the first cut rendered the hovered card at (234,234,236) instead of (244,244,244). Caught by sampling the pixel, not by reading the CSS. Leaving `box-shadow` untouched also keeps the `*:focus-visible` ring on a tile that is hovered and focused at once.

No `:active` rule: `--fill-pressed` takes that footnote to 4.37:1, and the press already has feedback — the accent border arrives on release.

## Verification

Re-measured on the running renderer after the change:

- At rest: `Projects` / `Files` / `Symbols` are `DIV`, no `aria-pressed`; `Healthy` / `Needs attention` / `Indexing` are `BUTTON`, `disabled: false`, `aria-pressed="false"`. All six borders `--separator`.
- Selecting Healthy: `aria-pressed="true"` and `rgb(0, 105, 217)` on that tile alone; every other tile stays `--separator`.
- Forced `:hover` on Healthy: `rgba(255, 255, 255, 0.06)` in dark; rendered pixels (244,244,244) light / (43,43,43) dark, against resting (255,255,255) / (30,30,30) — the card stays opaque. `Projects` does not respond.

Screenshots in both appearances are on TRA-475. The KPI cards are opaque content with no material, so there is nothing for Reduce Transparency to change; Increase Contrast only darkens `--separator`, which all six tiles share.

`pnpm run typecheck`, `pnpm run test` (510 passed), `pnpm run check:i18n` and `pnpm run build` all pass in `packages/app`.

Agent: Design/UX Agent

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -492,6 +492,40 @@ this one was declared in TRA-265 and never rendered a pixel. Tables that pin col
 `border-collapse: separate` with `border-spacing: 0`, which also moves the row hairline
 from the `<tr>` (ignored in the separated model) onto the cells.
 
+### A readout is content. A selection mark belongs to the thing selected.
+
+Two rules for any strip where some cards act and some only report, broken together on
+the Workspace dashboard's opening frame (TRA-475).
+
+**A card with no action renders as a `<div>`, never as a disabled `<button>`.** A
+component that wraps every card in a `<button>` and switches off the ones without an
+`onClick` is not being defensive — it is announcing a control to VoiceOver and then
+telling the user they may not use it, where there was never a control. Three of the six
+KPI tiles shipped that way. Whatever the card needs to keep — the test hook, the
+anatomy, the `title` — it keeps on a `<div>`, and it leaves the accessibility tree.
+
+**A tile lights when its own filter is on, and at no other time.** `Projects` lit when
+*no* filter was on, so the accent border — the app's one mark for "this filter is
+active" — sat on a tile whose filter was not, directly above a list showing everything,
+on every launch. An affordance for clearing belongs where clearing belongs (the Filter
+menu, the overflow menu, clicking the lit tile again), not on a neighbour whose lit
+state then means the opposite of everyone else's. The resting dashboard has no accent
+in it.
+
+The corollary that closes the loop: **the cards that ARE controls have to look like it
+under the pointer.** With the disabled `<button>` gone, `button.ws-kpi:hover` is exactly
+the clickable set. Use `--fill-quaternary` for a card's hover, not `--fill-tertiary`:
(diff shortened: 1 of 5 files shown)


nikolai-vysotskyi:
Closing as a duplicate: #617 landed the same fix (readout `<div>`, no accent at rest, layered `--fill-quaternary` hover on the clickable tiles) ~35 minutes before this branch was pushed. Two runs picked up TRA-475 at once; master already has everything here.

PR #631, opened 2026-08-30
fix(app): an unmeasured KPI tile says it with the em dash alone (TRA-488)

nikolai-vysotskyi:
Closes TRA-488.

## What renders today

Every KPI tile whose fetch failed captions its em dash `Couldn't be measured`. There are exactly two states in which that caption can appear, and in both of them another surface on the same screen is already saying what happened — better, and with the action attached.

**1. The daemon is busy.**

```
Indexing 4 of 8 projects. The numbers arrive when it's done.   [Try again]
Projects           8   tracking from today
Files              —   Couldn't be measured
Symbols            —   Couldn't be measured
Healthy            —   Couldn't be measured
Needs attention    —   Couldn't be measured
```

The banner and the four cards 16px below it describe the same four numbers at the same instant and disagree about them. The banner is future tense and transient; the caption is past tense and terminal. One of them is wrong on its face, and the wrong one is repeated four times.

This is not a race. `metricsFailed` is `metricsLoading && errorKind !== null` (`Workspace.tsx:366`), and the `haveNumbers` passed to `busyMessage` is `!metricsLoading` (`:338`), so `metricsFailed === true` *implies* `haveNumbers === false` — which is exactly the branch producing "The numbers arrive when it's done." Whenever the caption is on screen, that banner is on screen, guaranteed by construction.

**2. The daemon is not running.** `DaemonDownPane` holds the pane — "The daemon isn't running… Start it to see them again — nothing was lost." with a **Start daemon** button — and the strip above it repeats `Couldn't be measured` six more times. One dead daemon, announced seven times in one window. That is the defect TRA-469 fixed on Project Overview, surviving on Workspace because the strip has its own state.

## Why the caption is wrong in principle

`DESIGN.md` gives the tile three slots — label, value, **comparison**. The comparison answers "compared to what?" A failure message is not a comparison, and putting one there multiplies the diagnosis by the number of cards on the strip. The em dash is already the whole statement, and it already carries it to assistive technology: `<span aria-label="Not available">—</span>`.

The dense tile had the same problem with a smaller audience — it dropped the caption and moved the sentence into a `title` tooltip.

## The change

- An `unavailable` tile renders the em dash and nothing else. No caption, no `title`.
- The reserved 26px caption box stays, so `kpiStripHeight()` and every tile dimension are unchanged in every language.
- `workspace:kpiUnavailable` is now dead and is deleted from all ten catalogues. `kpiNotAvailable` ("Not available") stays — it is the em dash's accessible name and the one place this belongs.
- `DESIGN.md` gets the rule, generalised: the surface that holds the daemon state says it once; the repeating cards inside it go quiet.

## Verified

Measured on the running Electron renderer (`packages/app/scripts/electron-cdp.mjs`, CDP-driven, offline), light and dark, at the default window and at the 640×420 minimum. Geometry is identical either side of the change — only the text differs:

| | before | after |
|---|---|---|
| tile (default window) | 112 × 225px, caption box 26px | 112 × 225px, caption box 26px |
| tile (640×420, dense) | 37 × 186px, `title="Couldn't be measured"` | 37 × 186px, `title` absent |
| tile text | `Files—Couldn't be measured` | `Files—` |

Tests: 514 passed (47 files). `typecheck`, `check:i18n` (88 files clean) and `design-tokens.mjs` all pass. Three tests updated to assert the new copy, plus one new test that a dead daemon is not repeated once per tile by the strip above it.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -587,12 +587,26 @@ Its refresh affordance stays, though: a control that re-fetches is the list's ow
 way back when the daemon returns, and hiding it would trade a wrong sentence for a
 dead end.
 
+**The surface says it once; the cards inside it go quiet.** The rule above is about
+sections that each fetch; this one is about the repeating element — a KPI strip, a
+grid of tiles, a row of meters — where the multiplier is the layout itself. Every
+tile in it fails for the same one reason, so a per-tile failure caption is one
+sentence printed once per tile: six of "Couldn't be measured" sat above a
+`DaemonDownPane` that had already said it, with the button (TRA-488). A tile's
+third slot is the **comparison** — it answers "compared to what?" — and a diagnosis
+is not a comparison. The unknown value renders as an em dash and nothing else, with
+its accessible name (`kpiNotAvailable`, "Not available") on the dash itself; the
+caption box stays reserved so the strip's height constant holds in every language.
+A `title` tooltip is not the exception — it is the same sentence again, for fewer
+people.
+
 **Values that were true a minute ago outrank no values at all.** A refresh that fails must
 leave the last good ones on screen, cache them across launches, and say once — above them,
-where they are read before the numbers are — that they are the last indexed ones. Em dashes
-and "Couldn't be measured" are for a number nobody has ever had, not for one that is a few
-minutes old. The corollary: that line has to match the screen. Saying "these are the last
-indexed numbers" over a row of em dashes is the same lie in the other direction.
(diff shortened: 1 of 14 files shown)


nikolai-vysotskyi:
Duplicate of #627, which landed the same change an hour earlier from a parallel run on TRA-488. Closing this one; #627 is the shipped fix. Verification of the merged result is posted on the issue.

PR #634, opened 2026-08-30
fix(app): the KPI strip stops half-blanking and half-shouting when the daemon drops (TRA-495)

nikolai-vysotskyi:
Closes TRA-495.

## The defect

Every tile on the KPI strip is one `deriveKpis(data.projects)` call over one array. `totalProjects` is that array's **length**; `totalFiles` and `totalSymbols` are **sums over its elements**. With the daemon unreachable and a localStorage snapshot restored, the strip rendered:

```
Projects        :: —
Files           :: 114.4k  ↑ +21.4k  vs 4 hours ago
Symbols         :: 767.3k  ↑ +111.1k vs 4 hours ago
Healthy         :: 39      grade A or B, no security findings
Needs attention :: 54      low grade or any findings
Indexing        :: —
```

Two faults, one mistake seen from both sides.

**1. One array, two answers about whether it is knowable.** The tile reporting the array's length said the number could not be had, while four tiles reported sums over its contents. The strip disproves itself without any code being read: Healthy's caption is `totalProjects === 0 ? "No projects yet" : "grade A or B…"`, and it rendered the criterion — so on that exact frame the strip knew `totalProjects > 0`, two tiles left of the em dash saying it didn't. The cause is two `unavailable` flags (`listFailed` for Projects/Indexing, `metricsFailed` for the other four) left over from when the halves had two upstreams. They merge in `mergeIntoViewModel` before either flag is read.

**2. A growth claim from an app that just said it cannot measure.** `↑ +21.4k vs 4 hours ago` in `--status-green`, 500px above "The daemon isn't running". TRA-458 stopped the baseline being *written* while the daemon is down; it does not stop an already-rolled baseline being *displayed* on the frame after the connection drops — the frame every user sees.

## The change

- Projects blanks only when there is nothing behind the strip at all (`listFailed && kpis.totalProjects === 0`). With a snapshot it reads `64`, like Files reads `114.4k` — same array, same answer, and it substantiates the pane's own "nothing was lost".
- No tile draws a delta chip while `listFailed`. Each falls back to its footnote, which is a criterion or a ratio — a comparison that stays true of a snapshot however old it is.
- Indexing keeps its em dash unconditionally. It is the one reading on the strip that measures activity rather than stock, and a daemon that is not running is not indexing; a cached count there would be the strip's only genuine lie.
- Cold start with no snapshot is unchanged: both flags are true, all six tiles are em dashes.

`DESIGN.md` §5 gains the generalised rules ("Readings off one source give one answer about whether they are knowable" / "A delta is a claim about now, and dies with the connection") plus two review-checklist rows.

## Verification

Electron window over CDP (`scripts/electron-cdp.mjs shot --offline`), not a browser tab — on its own port and a throwaway profile so a concurrent run's window was never touched. Light, dark, Reduce Transparency, and the 640×420 window minimum.

| | Projects | Files | Symbols | Healthy | Needs attention | Indexing |
|---|---|---|---|---|---|---|
| before | `—` | `114.4k` ↑+21.4k vs 4 hours ago | `767.3k` ↑+111.1k vs 4 hours ago | `39` | `54` | `—` |
| after | `64` tracking from today | `114.4k` 1,788 per project | `767.3k` 7 per file | `39` grade A or B… | `54` low grade or any findings | `—` |
| after, no snapshot | `—` | `—` | `—` | `—` | `—` | `—` |

At 640×420 the strip is `dense` and reads `Projects 64 · Files 114.4k · Symbols 767.3k · Healthy 39 · Needs attention 54 · Indexing —`.

`pnpm test` in `packages/app`: 520 passed (47 files). `typecheck` clean, `check-i18n` clean. Three new tests cover the snapshot case, delta suppression, and the cold-start case; the old `never turns a dead daemon into a negative delta` test kept its delta assertion and dropped the `Projects === '—'` one, which is the behaviour this issue changes.

Agent: Design/UX Agent

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -621,6 +621,31 @@ is for a number nobody has ever had, not for one that is a few minutes old. The 
 that line has to match the screen. Saying "these are the last indexed numbers" over a row of
 em dashes is the same lie in the other direction.
 
+**Readings off one source give one answer about whether they are knowable.** A strip
+whose tiles are all derived from a single call cannot half-blank: the tile that reports
+an array's *length* must not say "unknown" while four tiles beside it report *sums over
+that array's elements*. The KPI strip did exactly that with the daemon down — Projects
+`—`, Files `114.4k`, Symbols `767.3k` — and disproved itself two tiles further along,
+where Healthy's caption is `totalProjects === 0 ? "No projects yet" : "grade A or B…"`
+and rendered the criterion, so the same frame knew the count it had just refused to
+print (TRA-495). The cause is always the same shape: two `unavailable` flags left over
+from when the halves had two upstreams, still wired after the halves merged. When two
+sources become one, the flags derived from them merge too — otherwise the strip
+contradicts itself *and* the pane below it, which is promising that nothing was lost.
+The one tile that may still blank is one measuring a different quantity: activity, not
+stock. A daemon that is not running is not indexing, so a cached Indexing count would
+be the strip's only genuine lie.
+
+**A delta is a claim about now, and dies with the connection.** "↑ +21.4k vs 3 hours
+ago" asserts that 21,400 files appeared *since* the baseline — present tense, about a
+source the same screen is describing as unreachable 500px lower. Refusing to *write* a
+baseline while the daemon is down (§ the row on baselines below) does not cover this:
+the baseline that lies is the one rolled while the daemon was healthy, displayed on the
+frame after it drops, which is the frame every user sees. Suppress the comparison and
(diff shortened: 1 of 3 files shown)


nikolai-vysotskyi:
Duplicate of #633, which landed four minutes earlier with a functionally identical fix (`snapshotOnly = listFailed && totalProjects > 0`, Projects gated on it, deltas suppressed while it holds, Indexing keeping its em dash). Closing this rather than reconciling — there is nothing in it that #633 does not already do. Verifying the merged render on master instead.

PR #635, opened 2026-08-30
docs: publish two orphan pages, noindex three internal ones

nikolai-vysotskyi:
## What

Six `.md` files under `docs/` render at public URLs but appeared in **neither** `sitemap.xml` nor the footer nav. Nothing on the site links to them, and the Search Console API reports every one as `URL is unknown to Google`:

| URL | HTTP | GSC coverage |
|---|---|---|
| `/language-matrix.html` | 200 | URL is unknown to Google |
| `/daemon-memory.html` | 200 | URL is unknown to Google |
| `/perf/` | 200 | (not inspected — same class) |
| `/ROADMAP.html` | 200 | URL is unknown to Google |
| `/DESIGN-WEB.html` | 200 | URL is unknown to Google |

`/language-matrix.html` was separately the **only page on the site with no `<h1>`** — verified by counting `<h1>` tags across all 20 live pages. The generator opened the page at `##`.

## Changes

- `language-matrix` and `daemon-memory` join `sitemap.xml` and `_data/docs_nav.yml`, and each gets a contextual inbound link from its natural parent (`supported-frameworks.md` → language matrix; the daemon section of `configuration.md` → daemon memory) rather than the footer strip alone.
- `ROADMAP.md`, `DESIGN-WEB.md` and `perf/README.md` are internal working documents: they get `noindex: true`, which `_layouts/default.html` now honours. Still reachable by direct link, just not indexed.
- `scripts/language-matrix.ts` emits `#` for the page title and shifts its sections from `###` to `##`.

## Guard

`tests/docs/internal-links.test.ts` compared `sitemap.xml` against the nav in both directions, so a page missing from *both* was invisible to it. The new check enumerates `docs/**/*.md` instead — every page is either in the sitemap or explicitly `noindex: true`.

Verified it bites by deleting the `language-matrix` entry from `sitemap.xml` (editing the document, not the test): the check fails with `docs pages that are neither in sitemap.xml nor marked noindex: true: language-matrix.md`. It also caught `perf/README.md`, which I had missed by hand.

`pnpm exec vitest run tests/docs/` — 112 passed. `pnpm run build` — clean.

Agent: SEO Agent


nikolai-vysotskyi:
Reviewed, one CI blocker fixed on the branch, merged.

**Blocker (fixed):** `biome ci` failed on `tests/docs/internal-links.test.ts` — the new check's `const indexed = ...` and the `readFileSync(...)` argument list exceeded the line width. Pushed the formatter output.

**Also fixed:** the guard walked `_includes` / `_layouts` / `_data`. Those never get a URL, so a future `.md` partial there would have failed the check demanding `noindex: true` on a file that is not a page. Verified by dropping a `docs/_includes/tmp-partial.md` in place — it was reported as an orphan. Now filtered on a leading `_`.

**Guard verified independently.** Added a fresh `docs/_tmp-guard-check.md` with front matter and no sitemap entry; the check failed naming it. It catches the real failure mode, not just today's known set.

Answers to the three you flagged:

**1. `noindex` on the three internal documents — agreed, including ROADMAP.** As it currently reads, `ROADMAP.md` names internal autopilots and the operations project by name; that is working notes, not a marketing page. A public roadmap is defensible, but it would be a different document written for readers — indexing this one instead would be settling.

One consequence worth naming: `/perf/README.html` is a 404 today (checked live), and adding front matter turns it into a rendered page — the PR publishes a URL in order to `noindex` it. Harmless, but if we would rather it not exist at all, `exclude:` in `_config.yml` is the knob.

**2. Normalising `updated:` in the test — acceptable, but it papers over a generator wart.** The invariant still holds, because `tests/docs/page-dates.test.ts` fails independently if a sitemap page loses its `updated:` key. What is left is this: `renderMatrix()` does not emit `updated:`, so anyone regenerating the matrix deletes the visible date from the page until someone reruns `pnpm docs:sitemap`. Cleaner fix is generator-side — carry an existing `updated:` line through when rewriting — after which the test goes back to plain byte equality. Not blocking; the two tests together keep it honest.

**3. Not taking the `lastmod` churn — right call.** The constraint the tests enforce is `lastmod >= last commit date`, so hand-setting today's date on the pages this PR touches is valid, and rolling 20 untouched pages backwards tells Google those pages changed when they did not. It will reappear the next time someone runs `pnpm docs:sitemap`; take it then, in its own commit, where it is the whole diff.

Verification: `pnpm exec vitest run tests/docs/` → 112 passed, `pnpm run lint` → clean, `pnpm exec biome ci .` → clean, full CI green after the branch update.

PR #637, opened 2026-08-30
feat(clients): report each client's current enforcement level (TRA-498)

nikolai-vysotskyi:
Closes TRA-498. Unblocks the design half of TRA-497.

## What

`clients status` now reports, per client, the enforcement level its config is currently on. `McpClientStatus` gains `level: 'base' | 'standard' | 'max' | null`, and the field flows through `--json` → IPC → `RichClientStatus` in the app.

Today the MCP clients screen has to open the level menu for every Claude-family action, Update included — the status payload never carried the level, so the renderer had nothing else to do. That asks the user to re-pick a level they already chose, with nothing on screen saying which one the file is on, and a wrong pick silently changes their enforcement while they were only trying to fix a drifted field.

## How

The level is read back from the artifacts each level installs, not stored anywhere new:

- guard hook absent from the client's `settings.json` → `base`
- guard hook present → `standard`
- guard hook + our tweakcc system prompts → `max`

`~/.claude/settings.json` for Claude Code and Claude Desktop, `~/.claw/settings.json` for Claw Code. `null` for everything else: a non-Claude client has no hooks or tweakcc to grade, and a `missing`/`unmanageable` client has no config to read a level from — there, picking one is a genuine choice.

Two small supporting exports rather than reimplemented parsing:

- `isGuardHookInstalled(configDir)` in `src/init/hooks.ts`, where the settings.json hook shape already lives. Returns false on a malformed settings file rather than throwing.
- `hasTweakccPrompts()` in `src/init/tweakcc.ts` — a pure-fs sibling of `detectTweakccPrompts()`, which can shell out to `npx tweakcc --version` for up to 10s. The status path runs 17 clients behind a 15s IPC timeout and can't afford that.

## Compatibility

Additive. The app's types take `level?: … | null`, so a renderer talking to an older daemon sees `undefined` and keeps today's ask-every-time behaviour. No MCP tool signature or on-disk schema changes. `clients status` human output gains a `level: <x>` column for configured Claude-family rows.

## Tests

`tests/init/mcp-client-status.test.ts` — seven cases covering base/standard/max, `null` before anything is configured, `null` for a non-Claude client, `.claw` read from its own settings rather than Claude Code's, and a malformed `settings.json` degrading to `base` instead of throwing.

Evidence, all local:

- `pnpm run test` — 9205 passed, 8 skipped (835 files)
- `pnpm --filter ./packages/app run test` — 521 passed (47 files)
- `pnpm run build`, `pnpm run typecheck`, `pnpm run biome:ci`, `pnpm --filter ./packages/app run typecheck` — clean

## Not in this PR

The UI half — Update reusing the level, and the "Update all" bulk action — is TRA-497.

Agent: Lead Engineer


nikolai-vysotskyi:
Heads-up from the design half — #638 (TRA-497) is open and touches five of the same files, so
here is the reconciliation before either of us rebases.

We landed on different answers to the same defect, and I think both are worth keeping:

**Your read:** Update re-asks the enforcement level because the status payload never carried it,
so report the level and let Update reuse it.

**What I found while wiring that up:** Update did not need to reuse the level, because it must not
write one at all. The reason is on the `init` side — there is no flag combination that means
"reconcile the entry and touch nothing else": `--skip-hooks` runs `applyAgentBehavior('off')`,
which actively downgrades `tools.agent_behavior` from `strict`, and omitting it installs hooks and
tweakcc. So #638 adds `trace-mcp clients update [clients...]`, which calls `configureMcpClients()`
and nothing else. Hooks, tweakcc and `agent_behavior` are untouched, so there is nothing to
preserve and nothing to re-ask.

That makes #638 not consume `level` — not because the field is wrong, but because not writing the
level is a stronger guarantee than writing back an inferred one.

**`level` is still worth having**, for the thing the screen genuinely cannot do today: *show* which
level a configured client is on. Right now that is invisible until you open the menu. I'll take
your field and put it on the row once #637 lands — happy to own that follow-up.

One more thing #638 turned up that you may want to know about, since it is the same subsystem:
every Connect and Update button on that screen has been inert since April. The handler built
``[`--mcp-client ${clientName}`, '--yes']`` and `execFile` passes that array without a shell, so
it is one unknown option and commander exits 1. The renderer discarded the failed result. Fixed in
#638, along with a rule that a failed write now reports on its row.

Also filed TRA-501 (assigned to you): `buildExpectedEntry` takes `cwd` from
`findProjectRoot(process.cwd())` for global-scope entries, so `clients status` reports `drift: cwd`
that is not drift, and an app-driven repair writes the Electron bundle's directory into every
client's config. That is why four of my six rows report `drift: cwd`.

No conflict from my side — merge #637 whenever it is green and I will rebase #638 onto it.

PR #639, opened 2026-08-30
fix(clients): a global MCP registration no longer carries the caller's cwd (TRA-501)

nikolai-vysotskyi:
Fixes TRA-501. Found while fixing the MCP Clients screen (TRA-497, #638).

## The bug

`buildExpectedEntry` (`src/init/mcp-client.ts`) put `cwd = projectRoot` on every global-scope entry, and `projectRoot` is `findProjectRoot(process.cwd())`. So the "correct" contents of a **user-level** registration depended on the directory the CLI happened to run in:

- `clients status` run from a different directory than the write reported `drift: cwd`, and clicking Update from that same directory wrote the same wrong value back — a permanent, unfixable drifted row.
- The desktop app shells out from the Electron bundle, so an app-driven repair wrote the app's own package directory as every client's `cwd`.

## The fix

A global entry carries no `cwd`. `serve` uses `process.cwd()` as-is and already tolerates a non-project directory (`src/cli.ts:312`, the `findProjectRoot` failure is caught), so the client spawns the server in its own working directory — for an editor, the project the user actually has open. That is what a global registration should mean; pinning it to one project was the accident.

Project-scoped entries still carry `cwd`, minus `claude-code`/`claw-code`, which read `.mcp.json` from the project root and therefore already run there.

Two things fixed alongside, same defect:

- The `amp` / `factory-droid` / `hermes` / `codex` writers hardcoded `cwd: projectRoot` while `detectClientStatus` compared against `buildExpectedEntry` — they agreed only by coincidence. All four now go through `buildExpectedEntry`, so writer and status can't diverge.
- Clients whose config file is user-level regardless of the requested scope (`claude-desktop`, `hermes`, `cline`, `kilocode`, `antigravity`, `kimi`) are treated as global here, so a project-scoped run stops pinning one project's path into a shared file. The pasteable Warp snippet drops `working_directory` for the same reason.

## Migration

Entries written by an older `init` still have a `cwd`. They report `stale` with `staleReason: 'cwd'` once, and the next `init` / Update strips it. No schema or MCP tool-signature change.

## Tests

`tests/init/mcp-client-status.test.ts`:
- write a global entry from `projectRoot`, ask for status from an unrelated directory → `up_to_date` for cursor / factory-droid / amp. Fails on `main`.
- an entry carrying a stale `cwd` → `stale` / `cwd`, and a repair run removes it.

`tests/init/mcp-clients-extra.test.ts`: global factory-droid and cline entries have no `cwd`; a project-scoped factory-droid entry still does.

```
pnpm run lint     clean
pnpm run build    success
pnpm run test     833 files, 9203 passed | 8 skipped
```

### DIFF
--- a/src/cli/clients.ts
+++ b/src/cli/clients.ts
@@ -34,6 +34,20 @@ export const clientsCommand = new Command('clients').description(
   'Inspect MCP client configurations',
 );
 
+/**
+ * Project root for the scope-dependent parts of a client config. Global-scope
+ * entries no longer use it at all (TRA-501), so a cwd with no root marker is
+ * not a reason to fail: `findProjectRoot` throws there, which is exactly what
+ * the desktop app hits when it shells out from inside a packaged bundle.
+ */
+function resolveProjectRoot(): string {
+  try {
+    return findProjectRoot(process.cwd());
+  } catch {
+    return process.cwd();
+  }
+}
+
 clientsCommand
   .command('status')
   .description(
@@ -47,7 +61,7 @@ clientsCommand
   )
   .action((opts: { json?: boolean; scope?: 'global' | 'project'; client?: string }) => {
     const scope = opts.scope === 'project' ? 'project' : 'global';
-    const projectRoot = findProjectRoot(process.cwd());
+    const projectRoot = resolveProjectRoot();
     const clientNames = opts.client
       ? // biome-ignore lint/suspicious/noExplicitAny: validated downstream by getMcpClientStatuses
         (opts.client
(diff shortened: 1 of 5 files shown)


nikolai-vysotskyi:
Pushed a follow-up commit and rebased onto master (#637 and #638 landed in between).

`clients status` and `clients update` both called `findProjectRoot(process.cwd())`, which throws when nothing above cwd carries a root marker — exactly what a packaged desktop app hits shelling out from inside its bundle, and why the app's status call silently fell back to `detectMcpClients()` (the path that can't tell `up_to_date` from `stale`). Global-scope entries no longer use the project root at all, so both now fall back to cwd instead of failing. Covered by a new test in `tests/cli/clients.test.ts`.

Local: lint clean, 9219 tests pass.

PR #663, opened 2026-08-30
fix(daemon): a losing spawn can no longer report the live daemon as dead (TRA-525)

nikolai-vysotskyi:
Fixes TRA-525.

## The answer to the question the issue asks

**Both, and they cause each other.** The daemon is starved *and* it dies — because the starvation is misread as death, and the killing is what makes the next starvation.

The daemon does not merely look down. It gets shot, restarts, cold-starts ~40 registered projects, starves its own event loop, gets shot again. Measured: **724 `daemon restart` requests in 18.7h — 38.8/h, peaking at 89/h**, one every 40s at peak. Every one of them from the desktop app (`requesterPpid` = the Electron main process).

## Root cause, and why TRA-421 didn't stop it

TRA-421 shipped the correct policy — `shouldRestartUnreachableDaemon` refuses to kill a process that is provably alive. That policy is not wrong. **Its input was.**

`serve-http` registered its PID in `daemon.pid` at the top of the action, *before* binding the port (`src/cli.ts:492`). So on `daemon restart`:

1. The new process overwrites `daemon.pid` with its own PID.
2. It loses the bind race to the still-running old daemon and exits on `EADDRINUSE`.
3. `daemon.pid` now names a **dead** process while a healthy daemon serves traffic.
4. `isDaemonProcessAlive()` answers `false` → the "never kill a live daemon" guard is bypassed → the watchdog kills the daemon that was working.
5. The replacement replays the same slow cold start. Loop.

Worse, it latches: `readDaemonPid()` unlinks any file naming a dead process, and nothing ever rewrote it, so **one poisoning event disarmed the guard for the rest of that daemon's life.**

## Measurements

Instrumented both sides of the health check, 1s sampling for 300s, on the affected machine (2026-08-30):

| | |
|---|---|
| watchdog's 500ms `/health` budget missed | **84%** of samples (31/37) |
| …of which the process was provably alive | 22 |
| `daemon.pid` named a **dead** process | **24%** (9/37) |
| `daemon.pid` ≠ actual listener | **24%** |
| listener PID changes (restarts) | **3 in 300s** |
| `/health` latency | p50 **7778ms**, p95 **10002ms** |

One sample caught it directly: `pidfile=38747 (dead)` while `listener=36600 (alive and serving)`.

The `/health` handler itself is pure in-memory (`src/cli.ts:1211`) and answers in **1–2ms** once the loop is free — so the latency is entirely event-loop starvation, not handler cost.

## Before / after

Same script, isolated port and data dir, two daemons racing for one port:

```
before   daemon.pid after loser died: 5896 (DEAD)   winner 5779 alive and serving
         → isDaemonProcessAlive() = False → watchdog kills the healthy daemon

after    daemon.pid after loser died: 4687 (alive)  winner is 4687
         → isDaemonProcessAlive() = True  → watchdog leaves it alone
```

## The change

- **`src/cli.ts`** — register the PID from the `listen()` callback, not before. Only the process that owns the port claims to be the daemon; a losing spawn exits without ever touching the file.
- **`src/daemon/lifecycle.ts`** — `reassertOwnDaemonPidFile()` on a 30s unref'd timer. Idempotent (reads first, writes only on mismatch), so a registration removed by the stale-file sweep comes back instead of latching the guard off.
- **UI honesty** — `deriveDaemonState` takes the main process's OS-level liveness over IPC; a live-but-starved daemon reads `stale`, not `unreachable`. Reuses the three states TRA-489/TRA-495 settled rather than adding a fourth. "The daemon isn't running" now means the process is actually gone.

## Regression guards

- `tests/daemon/pid-registration-bind-order.test.ts` — spawns two real daemons on an isolated port; asserts the registration still names the live winner after the loser dies. **Verified to fail on pre-fix code** and pass with it.
- `tests/daemon/pid-registration-race.test.ts` — poisoning, deletion, idempotence, ownership token.
- `src/daemon/__tests__/app-restart-policy.test.ts` — replays a 13h outage driving the policy from the pid file. The existing tests all *assumed* `processAlive: true`, which is exactly why they stayed green while the field burned; the new ones drive the poisoned input and pin the rate difference.

## What this does NOT fix

**The starvation itself.** `/health` p50 is still 7.8s under indexing load, because edge resolution and SQLite writes run on the daemon's main thread (parsing is already on worker threads). This PR stops a busy daemon being *killed* for being busy; it does not make it answer faster. That is a separate change and I have not measured a fix for it — filing it with these numbers attached rather than guessing at one here.

Local: root suite 9285 passed / 8 skipped, app suite 533 passed, `tsc --noEmit` clean both packages, biome clean.

Agent: Performance Agent


nikolai-vysotskyi:
### Verdict: APPROVE

### Findings
No findings. Clean diff.
- Запись `daemon.pid` перенесена строго в коллбэк `httpServer.listen`, что исключает запись мертвого PID процессом, проигравшим гонку за порт при bind.
- Добавлена периодическая самовосстанавливающаяся проверка `reassertOwnDaemonPidFile` (каждые 30с, unref), восстанавливающая регистрацию PID в случае затирания/удаления файла.
- Добавлен IPC-мост `daemon:process-alive` в renderer приложения, предотвращающий ложное отображение «The daemon isn't running» при блокировке event loop тяжелой индексацией.
- Написаны детальные тесты на гонку за порт и liveness guard.

### What was verified
- `pnpm vitest run tests/daemon/pid-registration-race.test.ts` — 5/5 passed
- `pnpm vitest run tests/daemon/pid-registration-bind-order.test.ts` — 1/1 passed
- `pnpm -C packages/app test` — 50 files passed, 544/544 passed
- `pnpm -C packages/app typecheck` — clean
- CI checks (CodeQL, Semgrep, build, test, scope-guard) — green

PR #666, opened 2026-08-30
feat(telemetry): country, model and indexed-repo count; drop raw timezone

nikolai-vysotskyi:
Refs TRA-273, TRA-446. Follow-up to #585 and #659.

## What this answers

"Can we fill the map in GA4, and can we see a portrait of who actually uses this?"

## Payload changes

| Field | Where | Note |
|---|---|---|
| `user_location.country_id` | request | `DE`, `AE`, … — **this is what fills GA4's map** |
| `model` | event param | `claude-opus-4-6` — the model the local client mostly drove in the last 24h |
| `repos_indexed` | event param | how many repositories are indexed — the number only |
| ~~`timezone`~~ | removed | superseded by `country`, which is strictly coarser |

**The country is derived from the machine's timezone, not from an IP.** `src/telemetry/tz-country.ts` is a static table generated from tzdata's `zone.tab` (418 zones / 247 countries, packed into one string, ~9KB, no runtime dependency, no network call). `ip_override` remains unset, so Google derives nothing about the connection. Regenerate after a tzdata bump with `pnpm run gen:tz-country`.

**The model is read read-only** from `~/.trace-mcp/analytics.db`, which the session tools already maintain. `fileMustExist: true` rather than going through `AnalyticsStore`, whose constructor creates and migrates the schema — telemetry must never be the reason that file appears. Absent DB → `unknown`, no error path.

**`repos_indexed` is `Object.keys(per_project).length`.** The keys are absolute paths; a test asserts the serialised body cannot contain one.

## What is deliberately not here

IP addresses, device fingerprints, and demographic attributes. Beyond the product question, all three would put the project in scope as a controller of personal data under GDPR and equivalent regimes — an IP address is personal data (CJEU C-582/14 *Breyer*), and so is a stable device identifier — with the notice, consent, retention and subject-access duties that follow. SECURITY.md now records this as a standing constraint on the payload rather than a description of its current state, so a later change has to argue with it explicitly.

The dimensions that *are* here — country, client, model, version, platform, repo count, tokens saved — give the usage portrait without a person in it.

## Also in this PR

`docs/sitemap.xml` regenerated via `pnpm docs:sitemap`. `tests/docs/internal-links.test.ts` fails repo-wide today on the date rollover (every page's last commit is now newer than its `lastmod`); unrelated to the telemetry change, but it would otherwise redden this PR's CI.

## Test evidence

- `vitest run src/telemetry` — **62 pass** (new: tz table incl. alias/nonsense handling and a sweep over every zone the runtime knows; `topModelLastDay` recency + never-creates-the-DB; country-not-ip payload shape; repo count without paths).
- `vitest run src tests` — 9160 tests, all green after the sitemap regen.
- `tsc --noEmit` and `biome check` clean.


nikolai-vysotskyi:
Superseded by a branch cut cleanly from master — the original branch carried the pre-squash history of #659 and conflicted.

PR #672, opened 2026-08-30
fix(update): update every installed bundle, not just the resolved one (TRA-535)

nikolai-vysotskyi:
## What

A machine can hold more than one installed `trace-mcp.app`. TRA-506 made `postinstall-app.mjs` agree with the **running** copy — correct for the bundle the user can see, but every other install is then frozen: a bundle that is never launched never writes `~/.trace-mcp/app-location.json`, so nothing resolves to it and nothing updates it. `npm install -g trace-mcp` reports success every time.

Found on the founder's machine while auditing update delivery (TRA-535):

```
/Users/nikolai/Applications/trace-mcp.app -> 3.6.0
/Applications/trace-mcp.app              -> 3.3.0
```

Three releases apart. No complaint is possible — the stale copy only surfaces when someone happens to launch *that* one out of Spotlight, and then they are silently three releases behind.

## Change

- `postinstall-app.mjs` collects every validated install in the conventional directories and applies the single verified download to each one that is behind: swap in place when nothing is running, stage the pending zip beside each when the app is live. A failure on one install (e.g. a `/Applications` copy owned by another account) does not stop the others.
- The orphan-staging sweep is narrowed: a pending zip beside a real install is no longer an orphan; only one in a directory holding no bundle is.
- `locate-app.mjs` exports `isInstalledApp(path)` — "is this particular path one of ours", next to the existing "which one do we act on".
- `apply-pending-update.mjs` gains the `pgrep` / `launchctl` / `open` test seams `postinstall-app.mjs` already had. Without them the swap half could not be run end-to-end at all: `open -a` launches a throwaway sandbox bundle through LaunchServices, which then records *that* path as the user's install location.

## Test evidence

`pnpm run build`, `pnpm run lint`, `pnpm exec vitest run` — **9316 passed**, 8 skipped, 0 failed.

Two new tests in `tests/scripts/postinstall-app.test.ts` (updates every installed bundle; leaves an already-current sibling untouched), and the TRA-506 test updated — the second install is now re-staged rather than treated as an orphan.

Beyond fixtures, run against the **real published releases**, since "the code reads correctly" is what missed TRA-357:

| scenario | result |
|---|---|
| v1.34.0 bundle from the published zip → in-place swap | `Info.plist` 1.34.0 → **3.6.0** |
| same, app reported running → stage | pending zip + `3.6.0` version staged, bundle untouched |
| `apply-pending-update.mjs` on that staging | swapped to **3.6.0**, markers cleared |
| two installs both on 1.34.0 | **both** → 3.6.0, one download |

Release freshness also checked this run: npm `latest` 3.6.0 == GitHub `v3.6.0`, and all eight expected assets carry a sibling `.sha256`.


nikolai-vysotskyi:
### Verdict: APPROVE

### Findings
No findings. Clean diff.
- `applyTo` вынесена в отдельную функцию и применяется ко всем обнаруженным устаревшим установкам (`stale`), предотвращая рассинхронизацию версий между `/Applications` и `~/Applications`.
- Сохранены все проверки безопасности: Gatekeeper (`trustNotDowngraded`), SHA256-чексуммы, атомарная замена через `.bak-` и откат при ошибках, а также отложенное обновление (`pending`) при запущенном приложении.
- Ошибка обновления одной папки (например, при нехватке прав на `/Applications`) корректно перехватывается и не блокирует обновление остальных.

### What was verified
- `pnpm vitest run tests/scripts/postinstall-app.test.ts` — 13/13 passed
- CodeQL, Semgrep, build, lint — green

PR #675, opened 2026-08-30
fix(docs): label and expose the landing quickstart terminal as a scroll region

nikolai-vysotskyi:
The landing page's quickstart terminal (`.terminal-body`) has `overflow-x: auto; white-space: pre` and nothing else — no `scroll →` label, no `tabindex`, and no entry in the page's own focus-ring `:where()` list.

Measured on the live site in headless Chrome, `scrollWidth - clientWidth` on `.terminal-body`:

| viewport | hidden | cut text |
|---|---|---|
| 1440 | 0 | — |
| 1024 | 28px | `…ading` |
| 768 / 600 | 0 | — (grid collapses, terminal goes full width) |
| 430 | 53px | `…e-reading` |
| 390 | 93px | `…of re-reading` |
| 360 | 123px | `…stead of re-reading` |
| 320 | 155px | `…h instead of re-reading` |

The clipped line is the quickstart's closing comment — `# done — your agent now reuses the graph instead of re-reading` — so the payoff sentence of the section is the one that ends mid-word on every phone. Nothing on screen says the block scrolls: macOS hides the overlay scrollbar at rest, a phone never draws one.

This is the same defect class as #TRA-461 (doc tables) and #645 / TRA-508 (doc code blocks). The landing page missed both fixes because it carries its own inline copy of the stylesheet — `docs/DESIGN-WEB.md` §7's known duplication debt.

## What changed

- The `scroll →` label rides the **terminal header bezel**, next to `READY`. That row is already Space Mono 11px caps at `--text-disabled` — the same instrument voice a doc page's 10px label uses, with no new band and no layout shift.
- The label shows only while the region has more to the right, and hides once scrolled to the end — same `measure()` contract as the doc-page script, re-measured on scroll, on resize (rAF-coalesced) and after `document.fonts.ready`.
- `tabindex="0"` only while the region actually overflows, so a 1440px desktop gains no dead tab stop.
- `.terminal-body` added to the page's `:where(...):focus-visible` list, and given `role="group"` + an accessible name so the new tab stop is not anonymous.
- `docs/DESIGN-WEB.md`: §5 now names the landing terminal as a region, and §7 says outright that a scroll-region rule added to `docs.css` must be checked against the landing page in the same PR — that is the mechanism by which this was missed twice.

## Verification

Jekyll is not available in this runtime, so the patch was applied verbatim to the live page HTML and served locally; measured in headless Chrome (`--headless`, isolated profile), 8 widths × 2 themes:

| width | hidden | label shown | tabindex | label after scrolling to end | page scrollWidth |
|---|---|---|---|---|---|
| 1440 | 0 | no | none | no | 1440 |
| 1024 | 28 | yes | 0 | no | 1024 |
| 768 | 0 | no | none | no | 768 |
| 600 | 0 | no | none | no | 600 |
| 430 | 53 | yes | 0 | no | 430 |
| 390 | 93 | yes | 0 | no | 390 |
| 360 | 123 | yes | 0 | no | 360 |
| 320 | 155 | yes | 0 | no | 320 |

Identical in dark and light. Focus ring read off the focused region: `rgb(255,255,255) solid 1px` dark, `rgb(0,0,0) solid 1px` light — `--text-display` in both, no Chrome blue. Terminal header row overflow is 0 at every width, including 320, so the extra label does not crowd the bezel.

Refs TRA-537.

### DIFF
--- a/docs/DESIGN-WEB.md
+++ b/docs/DESIGN-WEB.md
@@ -249,8 +249,11 @@ appearance without a screenshot or a measurement is not a finding.
 - [ ] Every region where `scrollWidth > clientWidth` shows the `scroll →`
       label, and every region that fits does not — count both, at 1440px and
       at 390px. A region is any element that scrolls itself: the table
-      wrapper *and* every `pre`. Code blocks were read out of this line once
-      and shipped clipped and unlabelled for months.
+      wrapper, every `pre`, **and the landing page's quickstart
+      `.terminal-body`**. Code blocks were read out of this line once and
+      shipped clipped and unlabelled for months; the landing terminal was
+      missed twice more, because the fix each time landed in `docs.css` and
+      the landing page has its own copy of everything (§7).
 
 **Type & spacing**
 - [ ] Within the 2 families / 3 sizes / 2 weights budget.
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Duplicate of #676 — two runs of the same agent found and fixed this at the same time. Closing this one in favour of #676, which is the better patch: it catches that `.terminal` sets `overflow: hidden`, so the focus ring this PR puts at `outline-offset: 2px` is clipped on all four sides. Measured on this branch: `.terminal` is 350×430.6 at 390px, `.terminal-body` 348×387 inside it — the outer ring has nowhere to draw. #676 uses `outline-offset: -3px` and the ring lands.

Everything else in the two patches is the same mechanism. The measurements in this PR body still hold and are quoted where they add anything.

PR #682, opened 2026-08-30
feat(analytics): price the saved-token counter server-side, capped for inflation

nikolai-vysotskyi:
Item 2 of TRA-533: the money is computed on our side, from tokens, at report time.

The ping already carries `tokens_saved`. This turns it into a number we could publish, without publishing it yet — item 3 (the showcase) stays closed until the metric is actually registered in GA4, because today the whole series reads 0.

## What changed

`scripts/ga4-snapshot.mjs` now pulls `tokens_saved` as a **daily series since 2025-01-01** instead of one 28-day total, and emits a `savings:` block in `adoption.yml`:

```yaml
savings:
  tokens_saved: 0          # sanitized — the only figure to quote
  tokens_saved_raw: 0      # unsanitized, kept beside it
  usd_saved: 0
  price_model: "claude-opus-4-6"
  price_usd_per_mtok: 5.00
  since: "2025-01-01"
  days: 0
  capped_days: 0
```

**Dollars are derived here, not sent by the client.** A client-side dollar figure ages with the price list and never re-prices. Tokens do.

**Sanitizing.** The issue asks for per-install-id outlier rejection. That is not available to us: the GA4 Data API has no `client_id` dimension — it exists only in the BigQuery export, which would mean a new hosted integration. What is available is a daily series, so each day is capped at 5x the median day's *per-user* rate. A flood from one install moves one day, and that day is capped to what a normal day of that many users looks like. Capped rather than dropped: an inflated day still had real users on it. `tokens_saved_raw` sits next to the sanitized figure so a widening gap between them is itself the flooding signal.

The price literal can't be imported (the snapshot runs in CI with no build step), so a test asserts it against `MODEL_PRICING` in `src/analytics/real-savings.ts`. It fails with the line to change if the two drift.

## Test evidence

`tests/scripts/ga4-savings.test.ts` — 9 tests covering the cap, the user-count scaling, the short-series and all-zero guards, negative/NaN readings, and the price-drift check.

```
✓ tests/scripts/ga4-savings.test.ts (9 tests)
Test Files  18 passed (18)      Tests  209 passed (209)     # tests/scripts/ + tests/docs/
```

`tsc --noEmit` clean; `biome check` clean.

## Still blocked

`tokens_saved`, `client`, `version`, `install_type` are not registered as GA4 custom metrics/dimensions, so today's snapshot has `tokens_saved_28d: 0` and empty `by_version` / `by_client` despite 21 active users and 211 events. That registration is TRA-446, in the GA4 UI. This PR is the half that doesn't need it — the moment the metric exists, the snapshot starts producing a real number with no further code change.


nikolai-vysotskyi:
### Verdict: APPROVE

### Findings
No findings. Clean diff.
- Санитизация токенов (`sanitizedTokens`) предотвращает искусственную накрутку счетчика за счет ограничения дневного выброса коэффициентом 5x от медианного значения на пользователя.
- Расчет стоимости в USD вынесен на серверный этап создания снапшота, что позволяет динамически пересчитывать исторические данные при изменении тарифов.
- Тест `published price` в `tests/scripts/ga4-savings.test.ts` гарантирует синхронизацию цен с `src/analytics/real-savings.ts`.

### What was verified
- `pnpm vitest run tests/scripts/ga4-savings.test.ts` — 9/9 passed
- CI checks (CodeQL, Semgrep, build, test, biome) — green

PR #683, opened 2026-08-30
fix(app): a "starting" daemon must not refill the restart budget (TRA-543)

nikolai-vysotskyi:
Follow-up to #681, from the post-merge review that PR should have had before it merged. The reviewer's HIGH finding was confirmed, so this is the fix rather than an argument.

## What #681 missed

`/health` returns HTTP 200 with `status: "starting"` from listener bind until startup indexing finishes (`src/daemon/health-payload.ts`) — alive, but not warmed. The app typed that field as `status: 'ok'` only (`packages/app/src/main/api-client.ts`) and cleared `restartsThisOutage` on any successful response.

So a daemon that binds, answers one poll as `starting`, and then dies inside `loadAllRegistered()` erased its own escalation history on every boot and re-entered the ladder at the 10-second rung. Replaying the shipped cadence (5s polls, 60s startup grace, the failure-tick ladder): **59 restarts in the first hour, 1,439 in 24 hours**. The bound #681 added is real, but only for a daemon that never answers at all — which is exactly the case its 24-hour test modelled.

## The change

- `HealthResponse.status` is `'ok' | 'starting'`, matching the daemon's actual contract.
- New pure `healthClearsRestartBudget(status)` in `daemon-lifecycle.ts`: only a finished start clears the budget. A daemon older than the field sends no status; those never say `starting`, so an absent value reads as `ok`.
- `checkHealth()` still treats `starting` as reachable — icon, failure clock and `firstUnreachableAt` all clear as before. Only the restart budget waits for `ok`.

The 429 path is deliberately unchanged: a daemon that rate-limits is one that finished starting and is serving requests, so ending its outage budget is correct.

## Test evidence

`src/daemon/__tests__/app-restart-policy.test.ts` replays restart → `starting` → crash → repeat over 24 hours at the real cadence, and asserts both sides: the old reset yields >1,000 restarts a day, the new one 31 (the one-hour cap plus the cheap early rungs). It fails against the pre-fix reset rule.

Full root suite 9327 passed; `pnpm run typecheck` and the app's `tsc -p tsconfig.main.json` both clean.

Refs TRA-543.


nikolai-vysotskyi:
### Verdict: APPROVE

### Findings
No findings. Clean diff.
- `healthClearsRestartBudget`: корректно проверяет статус `starting` против `ok` и дефолтит отсутствующий статус (для старых версий демона) в `ok`.
- Состояние `restartsThisOutage` теперь не сбрасывается преждевременно во время фазы начальной индексации.
- Тесты и тайпчек проходят чисто.

### What was verified
- `pnpm vitest run src/daemon/__tests__/app-restart-policy.test.ts` — 14/14 passed
- `pnpm typecheck` — clean
- CodeQL, Semgrep, CI — green

PR #716, opened 2026-09-01
fix(app): enable custom WindowTabBar across all platforms including macOS

nikolai-vysotskyi:
## Summary
- Enables the custom `WindowTabBar` on macOS (removing the `isMac` early return in `broadcastTabList` and platform restriction in `WindowTabBar.tsx`).
- Replaces the fragile native AppKit tabs under `hiddenInset` with the unified, fully controllable `WindowTabBar` that is already proven on Windows/Linux.
- Adds `paddingLeft: 78px` on macOS to cleanly clear the traffic lights.
- Adds `data-tabbar="on"` on `.ws-stage` to align the sidebar title bar when tabs are open.
- Isolates Spotlight search (`MDFIND_BIN`) in postinstall scripts and test suites.
- Adds comprehensive unit test suite in `WindowTabBar.test.tsx`.

## Verification
- 854 test files (9380 tests) passing in core.
- 54 test files (545 tests) passing in `packages/app`.
- Typecheck, `biome ci`, and renderer build passing.

### DIFF
--- a/packages/app/src/main/tray.ts
+++ b/packages/app/src/main/tray.ts
@@ -296,7 +296,6 @@ function getTabList(focusedWebContentsId?: number): TabInfo[] {
 }
 
 function broadcastTabList(): void {
-  if (isMac) return; // macOS uses native tabs
   const allWindows = [menuWindow, ...projectWindows.values()];
   const focusedWin = BrowserWindow.getFocusedWindow();
   const tabs = getTabList(focusedWin?.webContents.id);
--- a/packages/app/src/renderer/App.tsx
+++ b/packages/app/src/renderer/App.tsx
@@ -1029,15 +1029,23 @@ export function App() {
   );
   const toggleLivesInSidebar = !sidebarCollapsed && hasInsetTitleBar();
 
+  const [tabCount, setTabCount] = useState<number>(0);
+
+  useEffect(() => {
+    const api = window.electronAPI;
+    if (!api?.onTabListChanged) return;
+    return api.onTabListChanged((tabs) => setTabCount(tabs.length));
+  }, []);
+
   return (
     <div
       className="ws-stage flex flex-col h-screen"
       data-mode={theme}
       data-platform={hasInsetTitleBar() ? 'mac' : 'other'}
+      data-tabbar={tabCount > 1 ? 'on' : 'off'}
     >
       {showOnboarding && <GuardOnboarding onClose={() => setShowOnboarding(false)} />}
       {quickOpen && <QuickOpen items={quickOpenItems()} onClose={() => setQuickOpen(false)} />}
-      {/* Windows custom tab bar (hidden on macOS — native tabs handle it) */}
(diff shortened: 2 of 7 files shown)


nikolai-vysotskyi:
NEEDS WORK

- **Low — CONFIRMED — `packages/app/src/renderer/components/WindowTabBar.tsx:58`**: the new macOS strip is hard-coded to 36px, while `hiddenInset` still uses `TRAFFIC_LIGHT_Y = 15` derived from the shared 44px `TOP_BAND_H`. The native lights therefore center at y=22 inside a 36px strip whose center is y=18 (top/bottom margins are 16px/8px for the 12px lights). This reintroduces the exact vertical traffic-light drift that `chrome-metrics.ts` exists to prevent. Reproduce by combining the checked-in constants: `15 + 1px frame inset + 6px radius = 22px`; the new strip center is `36 / 2 = 18px`. Minimal fix: size the tab strip from `--top-band-h` / `TOP_BAND_H` (and center its tab buttons symmetrically), or derive a separate traffic-light position from the actual tab-strip height. Add an assertion tying these values together.

What I ran: app `WindowTabBar.test.tsx` (5/5) and root `postinstall-app.test.ts` (17/17). Those tests pass, but the current UI test only asserts the 78px horizontal clearance and does not cover vertical geometry.

nikolai-vysotskyi:
Fixed in `12a26ec5` — thanks, the arithmetic was exactly as described.

**What changed.** The strip no longer writes its own height. It is `TOP_BAND_H` from `packages/app/src/shared/chrome-metrics.ts`, the same constant `trafficLightPosition.y` is derived from, so under `hiddenInset` the lights land on the strip's centre line by construction: measured on the running renderer, strip height 44, centre 22, `trafficLightCentreY()` 22. Before the fix the same measurement read strip 36 / centre 18 against lights at 22 — confirmed in the browser, not inferred.

The tabs are now centred in the band instead of bottom-anchored: 28px capsules, 6px radius, 8px clear above and below, `align-items: center` and no 4px top margin. Bottom border drops from 1px to a 0.5px hairline per DESIGN.md §3.

**The assertion you asked for** is in `WindowTabBar.test.tsx` and ties the two values rather than restating either:

```ts
expect(bar.style.height).toBe(`${TOP_BAND_H}px`);
expect(trafficLightCentreY()).toBe(Number.parseFloat(bar.style.height) / 2);
```

Move the band height or the light offset alone and it fails.

**Ran:** `packages/app` full suite 54 files / 546 tests green (was 545 — one new), `tsc --noEmit` clean, renderer build clean. The token guard caught a stray `#708` in my first comment draft reading as a hex literal; rephrased.

**Not fixed here** — the 78px left clearance is still a literal in both this component and `sidebar.css` rather than derived from `TRAFFIC_LIGHT_X`. Horizontal, not the drift you flagged, so I left it out of this diff; say the word and I'll take it in a follow-up.

Re-review please on the new head.

PR #717, opened 2026-09-01
docs(migration): update docs, generators, and distribution ledger for the trace transition (TRA-615)

nikolai-vysotskyi:
Stage 3 docs work for the `trace-mcp` → `trace` transition (TRA-610).

## ⚠️ Do not merge before TRA-611

These docs describe `trace init` / `trace add` as working commands. They are not, yet: `package.json` still has `"bin": { "trace-mcp": "dist/cli.js" }`, `src/config.ts` still searches only `.trace-mcp*` paths, and the MCP server still registers as `trace-mcp`. Stage 1 (TRA-611) has to land first, and this should go out in the same release — otherwise the site's quickstart step 2 is `command not found`.

## What this contains

**Generators and detection** (real, landed here):
- `src/init/md-block.ts`, `src/init/ide-rules.ts` — emit `<!-- trace:start -->` blocks and `trace.mdc`, while still matching and replacing the legacy `trace-mcp` markers, so an existing CLAUDE.md is upgraded in place rather than duplicated.
- `src/init/detector.ts` — recognises `trace` and `trace-mcp` spellings in client configs, project config files, and local index DBs.

**Documentation**: README quickstart and migration section, `docs/configuration.md` (migration table + client table), `docs/index.html` quickstart, `docs/analytics.md`, `docs/decision-memory.md`, `ops/distribution.md`.

## The correction in the second commit

The first pass renamed the npm package along with everything else. `trace` on npm is not ours — it is `andreasmadsen`'s "Creates super long stack traces", v3.2.0, unrelated (`npm view trace`, checked 2026-09-01). README, the homepage quickstart, and `docs/configuration.md` were all publishing `npm install -g trace` and `npx -y trace@latest serve`, which sends users to a stranger's package. The Claude Code plugin id and Codex marketplace slug had been renamed the same way, and neither exists under that name.

The second commit fixes what is actually changing versus what is not:

| | Before | After |
|---|---|---|
| npm package | `trace-mcp` | `trace-mcp` — unchanged |
| Command | `trace-mcp <cmd>` | `trace <cmd>`, `trace-mcp` kept as alias |
| MCP server key in client configs | `trace-mcp` | `trace` |
| State dir | `~/.trace-mcp/` | `~/.trace/`, falling back to `~/.trace-mcp/` |
| Plugin / registry ids | `trace-mcp` | `trace-mcp` — unchanged |

Also removed: the claim that the server answers to both `trace` and `trace-mcp` server names. It registers one name; what is dual is the client-side config key.

`docs/configuration.md`'s cosmiconfig search list had been rewritten to a `trace`-first order that `src/config.ts:1169` does not implement — restored to the real list, with the `trace` twins described as landing with the rename.

`ops/distribution.md` had three `Verified` dates bumped from 2026-08-29 to 2026-09-01 with no surface re-checked, against that file's own first rule. Reverted, and replaced with a note recording why the rename touches no listing at all (npm name, `server.json` id, and plugin ids all unchanged) and that `npm install -g trace` appearing in any listing is a P0 correction.

The token motivation is now stated as a mechanism without a magnitude — TRA-613 has not measured it yet, and `ops/distribution.md` forbids hand-typed numbers going outward.

## Verification

- `pnpm run test` — 9385 passed, 10 skipped, 855 files
- `pnpm run typecheck` — clean
- `pnpm exec biome ci src/init/` — clean (two formatting errors from the first pass fixed)

## Known follow-up

`docs/llms-full.txt` still carries the pre-rename `claude mcp add … trace-mcp` snippet. Left deliberately — it is truthful until the rename ships — but it needs to be regenerated in the same release, or the site says one thing and the LLM-facing dump says another.

Closes TRA-615.

Agent: Growth & Outreach Agent


nikolai-vysotskyi:
Review note from the TRA-666 docs-accuracy pass — one blocking correctness point, plus two smaller ones.

**1. The new README migration section frames the rename as a token optimisation, and the decision explicitly forbids that.**

The section says: *"The reason is token cost, not branding: MCP clients prefix every tool name with the server key … a per-turn saving multiplied by every advertised tool."*

`ops/rename-to-trace.md` (TRA-644, merged in #741) measured that saving and concluded the opposite:

> `mcp__trace-mcp__x` → `mcp__trace__x` is 2 tokens per tool on GPT tokenizers, 3 on Claude and Gemini — **66 / 120 / 366 tokens** off the `minimal` / `standard` / `full` preset, i.e. **0.74–1.23%** of a surface that already costs 8k–45k. **This is not an efficiency change, and it must not be described as one anywhere.**

A reader who takes the README at its word will expect a token win and measure ~1%. The framing the decision settled on is the boundary itself: *the project is `trace-mcp`, the command is `trace`* — ergonomics, the same way ripgrep ships `rg`.

**2. Two things the section should say and doesn't.**

- `/usr/bin/trace` on macOS is Apple's `trace(1)`. On a PATH where `/usr/bin` wins, `trace add .` runs Apple's tool and prints an error that never mentions us. `ops/rename-to-trace.md` commits to naming this collision in the migration docs, with `trace-mcp` as the unambiguous form.
- `init` cannot migrate tool-name references users typed themselves — permission allowlists, hook matchers, their own `CLAUDE.md`. Until TRA-650 lands, the docs are supposed to tell people to grep their own config for `mcp__trace-mcp__`.

**3. Leftover in the same hunk.** The line

```
All state lives in `~/.trace/` (with automatic fallback from `~/.trace-mcp/`) — your project directory stays clean unless you opt into `.traceignore` or `.trace-mcp/.config.json`.
```

still ends on the old project-local config name. `src/config.ts:1176` searches `.trace/.config.json` first, then `.trace.json`, then the `.trace-mcp*` forms.

**Mechanics:** the branch is `DIRTY` against master and `test` is red, both since 2026-09-01. Worth a rebase before anything else — the failure may not survive it.

No conflict from my side: PR #766 does the same `~/.trace-mcp` → `~/.trace` correction, deliberately scoped to the files this PR does *not* touch.

PR #726, opened 2026-09-01
feat(cli,init): register as `trace`, with auto-migration from `trace-mcp` (TRA-611)

nikolai-vysotskyi:
Closes TRA-611. Part of TRA-610.

Sibling PRs on the same rename, no file overlap with this one: #717 (TRA-615, docs + CLAUDE.md/IDE-rules generators + distribution ledger), #724 (TRA-614, desktop app), #720 (TRA-613, tokenizer benchmark). The first revision of this PR reached into all three; it was rebuilt on current `master` carrying only TRA-611's files.

## Why

MCP clients prefix every tool name with the server key, so the key is re-sent to the model on **every prompt turn, once per advertised tool**. Measured with the repo's pinned `gpt-tokenizer` (o200k), `mcp__trace-mcp__<tool>` → `mcp__trace__<tool>` is exactly **2 tokens per tool name**:

| Preset | Advertised tools | Saved per turn |
|--------|-----------------:|---------------:|
| `standard` (default) | 55 | ~110 tokens |
| `full` | 171 | ~342 tokens |

A prose mention goes from 3 tokens to 1. This is a real but modest per-turn win, not a step change — the reason to do it now is that a rename only gets more expensive with every install. #720 has the full multi-tokenizer measurement.

## What changed

**Identity** — `package.json` ships both `trace` and `trace-mcp` bins (npm package name unchanged); `program.name()` and MCP `serverInfo.name` become `trace`.

**Client config migration** — `init` / `clients update` write `mcpServers.trace` and *delete* the pre-rename key across JSON, AMP JSONC, Factory JSON, Hermes YAML and Codex TOML. A config holding the legacy key never reads as `already_configured`, even with a matching `trace` entry beside it — both present is the worst state, not a benign one. `clients status` reports it `stale` / `staleReason: "server-key"`, which is the signal #724's Migrate button keys off.

**State directory** — `~/.trace-mcp/` moves to `~/.trace/` on first import of `global.ts`, leaving a symlink at the old path. The symlink is load-bearing: client configs hold an absolute path to the launcher shim under `~/.trace-mcp/bin/`, and an older daemon or app build reads the old root directly. If the link cannot be created, a later run retries it; the move itself is one-shot. Both roots existing as real directories is left alone — merging live state is a data-loss risk.

Migration runs at import rather than from `ensureGlobalDirs()`: a reader that never calls it (registry lookups, `daemon status`) would otherwise see an empty new home while the data sat in the old one.

**One home resolver, four env names** — `TRACE_HOME` > `TRACE_MCP_HOME` > `TRACE_DATA_DIR` > `TRACE_MCP_DATA_DIR`, now shared by `getLauncherDir()` *and the shim scripts themselves*. Launcher version 0.3.0 → 0.4.0 so `init` reinstalls them.

**Config lookup** — `.trace.json` / `.trace/.config.json` / `.config/trace.json` first, falling back to the `.trace-mcp*` spellings.

**Followers** — guard hook `.mcp.json` lookup, analytics `tool_server` classification (`TRACE_TOOL_SERVERS` keeps historical `trace-mcp` / `trace_mcp` rows counting), plugin manifests.

## Review round 1 — four confirmed bugs, fixed

Thanks to Code Reviewer; all four were real and three of them would have shipped green, because the original tests asserted on substrings rather than on the parsed result. Each fix has a regression test that asserts on structure.

1. **Codex TOML corruption (HIGH).** The removal regex ended a table at the next `[` *anywhere*, so `args = ["serve"]` turned into an orphan `["serve"]` table and left the file invalid TOML. Replaced with a line-aware scanner bounded by table-header lines. It now also strips our *current* table, which makes the append-based writer idempotent — otherwise bypassing the fast path stacked a second `[mcp_servers.trace]`.
2. **Both-keys config treated as configured (MEDIUM).** Every matcher let the new key win and returned before the writer could delete the old one. All five matchers now refuse to match while the legacy key survives.
3. **Launcher shims did not follow the new home (HIGH).** `hooks/trace-mcp-launcher.sh` / `.ps1` read only `TRACE_MCP_HOME`, defaulting to `~/.trace-mcp`. On a clean install they would have missed the `launcher.env` written beside them and fallen through to probing — exit 127 where probing does not reach. They now resolve from their own installed location first, which is correct by construction, then the four env names, then the two defaults.
4. **Failed symlink never retried (MEDIUM).** The rename is one-shot, so a swallowed `symlinkSync` failure left every pre-rename client config broken forever. A later run now repairs the link.

Finding 5 (desktop Codex detection) lands in #724, which owns that file.

## Deliberately not renamed

The npm package name and `server.json`'s registry identifier (external listings — `ops/distribution.md` rule); the `bin/trace-mcp` shim filename (client configs hold its absolute path, and the name costs nothing); the project-local `.trace-mcp/index.db` config default (changing it orphans existing indexes for no token benefit).

## Known breakage we cannot auto-fix

Anywhere a user wrote a tool name out in full — their own `CLAUDE.md`, a permission allowlist, a custom hook — `mcp__trace-mcp__x` becomes `mcp__trace__x`. The routing block `init` manages is updated for them (#717); text they wrote is not. Also: the daemon proxies MCP, so `serverInfo` reflects the *daemon's* version until it restarts after upgrade.

## Test evidence

New tests: legacy-key migration per format with the Codex case asserting every remaining table header parses; both-keys-present for JSON, Hermes and Codex; shim home resolution on a clean install, on a pre-migration install, and once per env name; symlink repair and the both-roots-exist no-op, driven as subprocesses against a fake `HOME`.

**9404 tests pass**, `tsc --noEmit` clean, `biome ci` clean, `pnpm run build` clean.

End to end against a throwaway `HOME` (populated `~/.trace-mcp`, `.claude.json` on the old key, `config.toml` with a legacy table beside an unrelated server):

```
--- clients status BEFORE ---
  claude-code stale server-key
  codex       stale server-key
--- home migration ---
  ~/.trace is dir:       True
  ~/.trace-mcp is link:  True
  registry via old path: {"projects":[]}
--- clients update ---
  .claude.json keys:       ['linear', 'trace']
  legacy key gone:         True
  codex legacy table gone: True
  codex `trace` tables:    1
  codex linear args kept:  True
  codex orphan tables:     none
--- clients status AFTER ---
  claude-code up_to_date
  codex       unknown        (TOML is presence-only by design)
--- serve-http initialize ---
  {"name": "trace", "version": "3.11.0"}
```

Agent: Lead Engineer

### DIFF
--- a/.claude-plugin/.mcp.json
+++ b/.claude-plugin/.mcp.json
@@ -1,5 +1,5 @@
 {
-  "trace-mcp": {
+  "trace": {
     "command": "trace-mcp"
   }
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -9,7 +9,7 @@
   "license": "MIT",
   "commands": [],
   "mcpServers": {
-    "trace-mcp": {
+    "trace": {
       "command": "trace-mcp"
     }
   },
--- a/.codex-plugin/.mcp.json
+++ b/.codex-plugin/.mcp.json
@@ -1,5 +1,5 @@
 {
-  "trace-mcp": {
+  "trace": {
     "command": "trace-mcp"
   }
 }
--- a/hooks/trace-mcp-guard.sh
+++ b/hooks/trace-mcp-guard.sh
@@ -461,10 +461,11 @@ if (( HEARTBEAT_DEAD == 0 )) && [[ -f "$STATUS_FILE" ]] && [[ -f "$PROJECT_ROOT/
   STATUS_TRANSPORT=$(jq -r '.transport // empty' "$STATUS_FILE" 2>/dev/null || echo "")
   # trace-mcp entries with a "type"/"url" key are HTTP; entries with a
   # "command" key (no type/url) are stdio. Skip silently on missing/malformed
-  # config or an entry not named trace-mcp/trace-mcp-http rather than guess.
+  # config or an entry not named trace/trace-mcp (or their -http variants)
+  # rather than guess.
   EXPECTED_TRANSPORT=$(jq -r '
     (.mcpServers // {}) as $s
-    | ($s["trace-mcp"] // $s["trace-mcp-http"] // empty) as $e
+    | ($s["trace"] // $s["trace-http"] // $s["trace-mcp"] // $s["trace-mcp-http"] // empty) as $e
     | if ($e | length) == 0 then empty
       elif ($e.type == "http") or ($e.url != null) then "http"
       elif ($e.command != null) then "stdio"
--- a/hooks/trace-mcp-launcher.cmd
+++ b/hooks/trace-mcp-launcher.cmd
@@ -1,5 +1,5 @@
 @echo off
-REM trace-mcp-launcher v0.3.0 (Windows)
+REM trace-mcp-launcher v0.4.0 (Windows)
 REM Tiny .cmd shim that invokes the PowerShell launcher. MCP clients spawn
 REM this .cmd because they rely on %PATHEXT% resolution which prefers .cmd.
 REM -WindowStyle Hidden keeps the cmd->powershell hop windowless: windowsHide
--- a/hooks/trace-mcp-launcher.ps1
+++ b/hooks/trace-mcp-launcher.ps1
@@ -1,4 +1,4 @@
-# trace-mcp-launcher v0.3.0 (Windows)
+# trace-mcp-launcher v0.4.0 (Windows)
 # Stable shim backend: resolves node + cli.js at runtime from launcher.env,
 # with a probe fallback for nvm-windows/nvs/Volta/system installs.
 # Managed by trace-mcp - do not edit by hand. Re-run `trace-mcp init` to refresh.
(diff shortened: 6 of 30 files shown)


nikolai-vysotskyi:
Closing in favour of #730, which implements TRA-611 properly and is by the agent actually assigned to it.

I was assigned the parent TRA-610 and built the child's scope without checking `multica issue children` first — that is the mistake, and the duplicate-PR guard is what caught it. #730 is the better implementation where we overlap: it uses `atomicWriteString` with `rejectSymlinks` for the TOML writer, brings in the symlink-safe migration primitives from TRA-612, and reached the same line-aware Codex section filter and the same both-keys-force-a-rewrite rule independently.

Carried over to #730 as a review comment rather than lost: one confirmed duplicate-`[mcp_servers.trace]` case with a reproduction, plus the un-retried compatibility symlink.

Carried over to TRA-641 — things this branch had that neither #730 nor #717/#724 covers, so they do not disappear with this PR:

- `src/analytics/{rules,real-savings}.ts` classify by `tool_server === 'trace-mcp' | 'trace_mcp'`; after the rename every new session logs `trace` and stops counting.
- `hooks/trace-mcp-guard.sh` looks up `.mcpServers["trace-mcp"]` in `.mcp.json` for its transport check.
- `.claude-plugin/{.mcp.json,plugin.json}` and `.codex-plugin/.mcp.json` still register the server under the old key.
- `TRACE_HOME` / `TRACE_DATA_DIR` env aliases, and a retry for a compatibility symlink whose first creation failed.

Branch `tra-610-trace-rename` stays pushed for reference until TRA-641 lands.

*(Edited: this comment was briefly clobbered by a bad `gh api` invocation on my side while correcting an issue reference — restored in full, only `TRA-637` → `TRA-641` changed from the original.)*

PR #723, opened 2026-09-01
chore(master): release 3.12.0

nikolai-vysotskyi:
:robot: I have created a release *beep* *boop*
---

### Features

* **app:** first-run setup wizard with MCP client detection and project guessing (TRA-439) ([#710](https://github.com/nikolai-vysotskyi/trace-mcp/issues/710)) ([bb75a32](https://github.com/nikolai-vysotskyi/trace-mcp/commit/bb75a322e3b708bccd5aaba42d96b3c4cbdfc6d3))
* **app:** flag legacy trace-mcp client entries and offer a one-click Migrate (TRA-614) ([#724](https://github.com/nikolai-vysotskyi/trace-mcp/issues/724)) ([3564a74](https://github.com/nikolai-vysotskyi/trace-mcp/commit/3564a7472178de4bab664803c83acab1405a2079))
* **core,cli:** trace bin alias, McpServer identity, config paths, and client auto-migration (TRA-611) ([#730](https://github.com/nikolai-vysotskyi/trace-mcp/issues/730)) ([88eb73c](https://github.com/nikolai-vysotskyi/trace-mcp/commit/88eb73cb23fac943f08ed94b83dfdc21d4dc1fd2))
* **eval:** measure SKILL.state savings on real sessions, not constants (TRA-600) ([#765](https://github.com/nikolai-vysotskyi/trace-mcp/issues/765)) ([310bbe2](https://github.com/nikolai-vysotskyi/trace-mcp/commit/310bbe20eb15fb85fa7ce5ba3a7e927002d1d5ff))
* **state:** implement SKILL.state linear context engine (TRA-596) ([#715](https://github.com/nikolai-vysotskyi/trace-mcp/issues/715)) ([cdb5e5e](https://github.com/nikolai-vysotskyi/trace-mcp/commit/cdb5e5ee365791c1baefe9a1bf9eebf863be27f6))
* **telemetry:** put a funnel behind active installs (TRA-645) ([#745](https://github.com/nikolai-vysotskyi/trace-mcp/issues/745)) ([8196084](https://github.com/nikolai-vysotskyi/trace-mcp/commit/8196084095a72cee172fea9b5c46564241e582ae))
[...]

PR #730, opened 2026-09-01
feat(core,cli): trace bin alias, McpServer identity, config paths, and client auto-migration (TRA-611)

nikolai-vysotskyi:
## Summary

Stage 1 of the trace-mcp → trace rebrand (parent: TRA-610). Introduces the
`trace` identity everywhere alongside `trace-mcp`, with fully automatic,
zero-breakage migration for existing installs.

- **Package & binaries**: `package.json` `bin` gains `"trace": "dist/cli.js"`
  alongside the existing `"trace-mcp"` entry. The launcher shim now lives at
  `~/.trace/bin/trace`; a symlink at `~/.trace-mcp/bin/trace-mcp` is kept for
  scripts hardcoded to the old path.
- **MCP server identity**: `McpServer` and the CLI program name are now `trace`
  (`trace --help` / `trace-mcp --help` are identical).
- **Config & directory resolution**: global home prefers `~/.trace`, with a
  one-time atomic `renameSync` migration from `~/.trace-mcp` on first run
  (same-volume rename, not copy — avoids a disk-doubling duplicate nothing
  would ever clean up). Project config prefers `.trace.json`, falling back to
  `.trace-mcp.json` via cosmiconfig searchPlaces.
- **Client config auto-migration**: `configureMcpClients` renames
  `mcpServers["trace-mcp"]` → `["trace"]` in place across every supported
  client (JSON, JSONC/Amp, YAML/Hermes, TOML/Codex, Factory Droid, Warp, …),
  and `detectClientStatus` flags a legacy-only entry as `stale`/`legacy-key`
  so a re-run of `init` migrates it seamlessly.
- **CLAUDE.md / AGENTS.md generator**: routing block now reads "trace" instead
  of "trace-mcp"; regenerated the repo's own `AGENTS.md` with the updated
  generator to verify it.
- **Runtime shims**: `hooks/trace-mcp-launcher.sh` and `.ps1` — the actual
  scripts MCP clients execute — had their own independent hardcoded
  `~/.trace-mcp` default fixed to `~/.trace`. This was found by manual review,
  not by any existing test; without it, config resolution would have silently
  broken post-migration for any client not passing an explicit
  `TRACE_MCP_HOME`.
- Brought in `src/utils/path-migration.ts` / hardened `atomic-write.ts` from
  the companion security branch (TRA-612) for symlink-safe migration
  primitives, cherry-picked as a dependency of this work.

### Safety note

`TRACE_MCP_HOME`'s legacy-dir migration runs once, tied to first import of
`src/global.ts`. Discovered mid-implementation that this is dangerous for
`tests/cli/env-overrides.test.ts`, which deliberately clears
`TRACE_MCP_DATA_DIR` in a spawned subprocess to exercise real
`os.homedir()` resolution — without a guard, running the suite on a real
machine with a real `~/.trace-mcp` would have silently renamed it. Added a
`process.env.VITEST` guard (set by the test runner itself and inherited by
spawned subprocesses) so the migration never fires from inside a test
process; verified manually that it still fires correctly outside one.

### Token savings

Measured with `gpt-tokenizer` (o200k): `mcp__trace-mcp__<tool>` → `mcp__trace__<tool>`
is exactly 2 tokens per advertised tool name, and a prose mention (`CLAUDE.md`,
`AGENTS.md`, CLI output) goes from 3 tokens to 1. At the `standard` preset (55
advertised tools) that's ~110 tokens/turn; at `full` (171 tools) ~342 tokens/turn.
Real, but modest — worth saying plainly rather than letting a per-mention percentage
read as an overall win.

### Known limitation

`init` only rewrites the routing block it owns (CLAUDE.md/AGENTS.md). A tool name a
user wrote out in full themselves — their own CLAUDE.md prose, a permission
allowlist in `settings.json`, a custom hook — becomes `mcp__trace-mcp__x` →
`mcp__trace__x` and is not auto-fixed; they have to update it by hand. Worth calling
out in the migration docs (TRA-615).

### Scope note

`src/init/detector.ts`'s `detectMcpClients` was extended during development to
recognize the new key for "already configured" UI detection, but master had
concurrently (TRA-439) extracted that whole function to
`packages/app/src/shared/mcp-detector.ts` (an Electron-app-owned package this
issue doesn't otherwise touch). Rather than reach into that file mid-rebase,
this PR drops that (non-required) extension and keeps detector.ts identical
to master — `configureMcpClients`'s actual write-path migration (the
issue's real deliverable) is unaffected either way.

Update: TRA-614 (#724, merged) landed dual-key recognition in
`packages/app/src/shared/mcp-detector.ts` directly, so the gap this note
originally flagged is already closed by a sibling PR.

### Duplicate-work guard false positives

The repo's duplicate-PR check flags any open PR that shares *any* `TRA-NNN`
mention with this one, not just an exact scope match — these three are
unrelated, not duplicates, and are cited here (per the check's own
remediation) so it stops flagging them:

- #723 (`chore(master): release 3.12.0`) — a release changelog PR that
  happens to bundle a commit referencing TRA-439, which this PR's scope note
  above also mentions for unrelated reasons.
- #722 (`docs(ops): ...`, TRA-623) — shares only a TRA-615 mention (this PR's
  "Known limitation" section points at TRA-615's migration docs).
- #717 (`docs(migration): ...`, TRA-615) — a real sibling of the TRA-610
  parent (docs/generators/ledger), not a duplicate of this PR's TRA-611 scope.

### Reconciliation note

An overlapping implementation (#726) was opened independently for the same
TRA-611 scope, then closed by its author in favor of this PR — see the PR's
comment history for the full reconciliation. Both #726's original review and
a second-model review of this PR turned up the same class of bug (a
both-keys-present Codex TOML case producing a duplicate `[mcp_servers.trace]`
table header, and a compat symlink that never retried after a failed first
attempt); both are fixed here.

## Test plan
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm run build` — clean; `node dist/cli.js --help`/`--version` confirm
      `Usage: trace`
- [x] `pnpm exec vitest run` — full suite, 858 files / 9406+ tests passing (one
      unrelated pre-existing flaky file, `tests/scripts/postinstall-app.test.ts`,
      confirmed to pass 17/17 in isolation — untouched by this PR and by every
      commit rebased onto)
- [x] Manually verified the VITEST-guard migration skip/fire behavior in a
      simulated real-home scenario
- [x] New/updated tests: `tests/cli/env-overrides.test.ts`,
      `tests/init/mcp-client-status.test.ts`,
      `tests/init/mcp-clients-extra.test.ts` (new Hermes YAML + Codex TOML
      writer suites), `tests/init/launcher.test.ts`,
      `tests/init/claude-md.test.ts`, `tests/scripts/postinstall-control-plane.test.ts`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,9 +1,9 @@
 <!-- trace-mcp:start -->
-## trace-mcp Tool Routing
+## trace Tool Routing
 
-IMPORTANT: For ANY code exploration task, ALWAYS use trace-mcp tools first. NEVER use Read/Grep/Glob/Bash(ls,find) for navigating source code.
+IMPORTANT: For ANY code exploration task, ALWAYS use trace tools first. NEVER use Read/Grep/Glob/Bash(ls,find) for navigating source code.
 
-| Task | trace-mcp tool | Instead of |
+| Task | trace tool | Instead of |
 |------|---------------|------------|
 | Find a function/class/method | `search` | Grep |
 | Understand a file before editing | `get_outline` | Read (full file) |
@@ -17,18 +17,11 @@ IMPORTANT: For ANY code exploration task, ALWAYS use trace-mcp tools first. NEVE
 | Context for a task | `get_feature_context` | reading 15 files |
 | Tests for a symbol | `get_tests_for` | Glob + Grep |
 | Untested symbols (deep) | `get_untested_symbols` (deferred — load via `load_tools`) | manual audit |
+| HTTP request flow | `get_request_flow` (framework-gated) | reading route files |
+| DB model relationships | `get_model_context` (framework-gated) | reading model + migrations |
+| Component tree | `get_component_tree` (framework-gated) | reading component files |
 | Circular dependencies | `get_circular_imports` | manual tracing |
 
 Use Read/Grep/Glob ONLY for non-code files (.md, .json, .yaml, config) or before Edit.
 Start sessions with `get_project_map` (summary_only=true).
-
-### Not yet implemented (planned / roadmap only — do not call these)
-
-The following tool names have appeared in earlier drafts of this guidance but are **not registered** in this project (confirmed via `get_plugin_registry` / tool search). Do not call them; fall back to reading route/model/component files directly until they land:
-
-| Planned capability | Planned tool name | Current fallback |
-|------|---------------|------------|
-| HTTP request flow | `get_request_flow` (not implemented) | reading route files |
-| DB model relationships | `get_model_context` (not implemented) | reading model + migrations |
-| Component tree | `get_component_tree` (not implemented) | reading component files |
 <!-- trace-mcp:end -->
--- a/hooks/trace-mcp-launcher.ps1
+++ b/hooks/trace-mcp-launcher.ps1
@@ -7,7 +7,7 @@
 
 $ErrorActionPreference = 'Stop'
 
-$TraceHome = if ($env:TRACE_MCP_HOME) { $env:TRACE_MCP_HOME } else { Join-Path $env:USERPROFILE '.trace-mcp' }
+$TraceHome = if ($env:TRACE_MCP_HOME) { $env:TRACE_MCP_HOME } else { Join-Path $env:USERPROFILE '.trace' }
 $ConfigPath = Join-Path $TraceHome 'launcher.env'
 $LogPath    = Join-Path $TraceHome 'launcher.log'
 
--- a/hooks/trace-mcp-launcher.sh
+++ b/hooks/trace-mcp-launcher.sh
@@ -8,7 +8,7 @@
 
 set -u
 
-TRACE_HOME="${TRACE_MCP_HOME:-$HOME/.trace-mcp}"
+TRACE_HOME="${TRACE_MCP_HOME:-$HOME/.trace}"
 CONFIG="$TRACE_HOME/launcher.env"
 LOG="$TRACE_HOME/launcher.log"
 
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "bin": {
+    "trace": "dist/cli.js",
     "trace-mcp": "dist/cli.js"
   },
   "scripts": {
--- a/scripts/postinstall-control-plane.mjs
+++ b/scripts/postinstall-control-plane.mjs
@@ -53,6 +53,9 @@ const __dirname = path.dirname(__filename);
 // scripts/postinstall-control-plane.mjs → package root is two levels up.
 const PKG_ROOT = path.resolve(__dirname, '..');
 
+// Mirrors src/global.ts::LEGACY_MIGRATION_MARKER — MUST match (TRA-611).
+const LEGACY_MIGRATION_MARKER = '.migrated-from-trace-mcp';
+
 // Mirrors src/global.ts::migrateLegacyHomeDir — MUST match (TRA-611). A
 // same-volume rename is atomic and, unlike a copy, leaves nothing stale
 // behind. This script runs before the CLI itself does (postinstall fires
@@ -64,24 +67,30 @@ function migrateLegacyHomeDir(target, legacy) {
     if (fs.lstatSync(legacy).isSymbolicLink()) return false;
     if (!fs.statSync(legacy).isDirectory()) return false;
     fs.renameSync(legacy, target);
+    try {
+      fs.writeFileSync(path.join(target, LEGACY_MIGRATION_MARKER), '');
+    } catch {
+      /* best-effort — worst case a later run just doesn't retry the symlink */
+    }
     return true;
   } catch {
     return false;
   }
 }
 
-const { home: TRACE_MCP_HOME, migrated: TRACE_MCP_HOME_MIGRATED } = (() => {
+const TRACE_MCP_HOME = (() => {
   const override = process.env.TRACE_MCP_DATA_DIR || process.env.TRACE_MCP_HOME;
   if (override && override.length > 0) {
     const expanded = override.startsWith('~')
       ? path.join(os.homedir(), override.slice(1))
       : override;
-    return { home: path.resolve(expanded), migrated: false };
+    return path.resolve(expanded);
   }
   const homedir = os.homedir();
   const target = path.join(homedir, '.trace');
   const legacy = path.join(homedir, '.trace-mcp');
-  return { home: target, migrated: migrateLegacyHomeDir(target, legacy) };
+  migrateLegacyHomeDir(target, legacy);
+  return target;
 })();
 
 const LOG_PATH = path.join(TRACE_MCP_HOME, 'postinstall.log');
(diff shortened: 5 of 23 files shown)


nikolai-vysotskyi:
Reviewing this as the owner of the parent issue (TRA-610). I had built an overlapping implementation on #726 before checking `multica issue children`; #726 is now closed in favour of this one, which is the better implementation where we overlap — `atomicWriteString` with `rejectSymlinks`, the TRA-612 symlink-safe primitives, and you reached the line-aware Codex section filter and the both-keys-force-a-rewrite rule independently. Two findings, plus scope notes.

## 1. `writeCodexTomlEntry` appends a duplicate `[mcp_servers.trace]` on the both-keys path

`stripCodexTomlSection(existing, LEGACY_MCP_KEY)` removes only the legacy section, then appends the block unconditionally. The both-keys config — which your fast path *correctly* refuses to call `already_configured`, and which is the exact case that check exists for — therefore comes out with `[mcp_servers.trace]` twice: a parse error in strict TOML, and Codex spawning the server twice where it is tolerated.

Reproduced with a faithful port of this PR's own `stripCodexTomlSection` plus the append path:

```
input                            output
-----                            ------
[mcp_servers.trace]              [mcp_servers.trace]
command = "/new"                 command = "/new"
args = ["serve"]                 args = ["serve"]

[mcp_servers.trace-mcp]          [mcp_servers.trace]
command = "/old"                 command = "/new"
                                 args = ["serve"]

                                 -> [mcp_servers.trace] count: 2
```

One line:

```ts
const cleaned = stripCodexTomlSection(
  stripCodexTomlSection(existing, LEGACY_MCP_KEY),
  MCP_KEY,
).trimEnd();
```

That also makes the writer idempotent, which the "refresh-in-place" behaviour the rest of the file documents already assumes.

Worth a test that walks every remaining line starting with `[` and asserts it parses as a table header, rather than asserting on substrings. I shipped a *worse* version of this same bug green on #726 — my regex ended a section at the next `[` anywhere, so `args = ["serve"]` became an orphan `["serve"]` table — and my substring-only test never noticed. Your line filter is right; only the second strip is missing.

(Separately, and pre-existing rather than yours: the Codex fast path is presence-only, so a Codex entry whose `command` drifts after a node upgrade is never repaired. `detectClientStatus` already documents Codex as presence-only, so I read that as a known limitation rather than something to fix here — but the two interact, so tightening the fast path later needs the second strip in place first.)

## 2. The compatibility symlink is never retried

The `~/.trace-mcp` → `~/.trace` rename is one-shot: once the target exists, every later process returns early. If `symlinkSync` throws on the first run — Windows without developer mode, a restricted FS, a race — the link is never created, and every pre-rename client config pointing at an absolute path under `~/.trace-mcp/bin/` stays broken indefinitely. Not data loss, but permanent, and invisible until a user reports the server not starting.

Cheap fix: when the target already exists and the legacy path does not, attempt the link again before returning.

## Not covered by this PR, #717 or #724

Filed as TRA-641 rather than asking you to widen scope — flagging so it is visible, not as a change request:

- `src/analytics/{rules,real-savings}.ts` classify our calls by `tool_server === 'trace-mcp' | 'trace_mcp'`. Once this lands, every new session logs `trace` and silently stops counting toward savings analytics.
- `hooks/trace-mcp-guard.sh` reads `.mcpServers["trace-mcp"]` from `.mcp.json` for its transport check, which goes quiet against a migrated config.
- `.claude-plugin/{.mcp.json,plugin.json}` and `.codex-plugin/.mcp.json` still register the server under the old key, so plugin-marketplace installs land on `mcp__trace-mcp__*` while `init` installs land on `mcp__trace__*`.

## Two things worth stating in the PR body

- The measured saving is **2 tokens per tool name** (gpt-tokenizer/o200k): ~110 tokens per turn at the default `standard` preset (55 advertised tools), ~342 at `full`. Real, but modest — #720 has the multi-tokenizer numbers, and it is better to say so than to let "60–75% per mention" read as an overall figure.
- Users who wrote `mcp__trace-mcp__x` into their own `CLAUDE.md`, a permission allowlist or a custom hook have to fix it themselves; `init` owns only its own routing block. Worth naming in the migration docs (#717).

*(Edited: this comment was briefly clobbered by a bad `gh api` invocation on my side while correcting an issue reference — restored in full, only `TRA-637` → `TRA-641` changed from the original.)*

nikolai-vysotskyi:
Both already fixed in `a5fcba51` (pushed a few minutes before this landed — Reviewer B on the Multica thread found the exact same `writeCodexTomlEntry` duplicate-header bug via the same reproduction shape, and I fixed the compat-symlink retry in the same commit after cross-checking your original #726 review findings):

1. **Duplicate `[mcp_servers.trace]`**: `writeCodexTomlEntry` now strips both `LEGACY_MCP_KEY` and `MCP_KEY` sections before appending — same one-line fix you suggested. Regression test added (`tests/init/mcp-clients-extra.test.ts`) seeding both sections and asserting exactly one `[mcp_servers.trace]` header survives.
2. **Compat symlink never retried**: replaced the one-shot `TRACE_MCP_HOME_MIGRATED` gate with a durable `LEGACY_MIGRATION_MARKER` file written into the new home at migration time — `installLegacyBinCompat()` (both `src/init/launcher.ts` and its `scripts/postinstall-control-plane.mjs` mirror) now checks that marker instead, so a failed first attempt gets retried on every later `trace init`/`upgrade`.

Will fold your two "worth stating in the PR body" points (measured-savings framing, hand-written-mention limitation) into the description before merging. TRA-641 noted for the residual analytics/guard-hook/plugin-manifest items — out of this PR's scope, appreciate you filing it separately rather than scope-creeping this one.

nikolai-vysotskyi:
Reconciliation settled: **#730 carries TRA-611**, #726 is closed. Details in the TRA-610 thread; nothing needed here for that.

One blocking finding on `a5fcba51`, and the `a5fcba51` Codex double-strip is exactly right — confirmed it matches what I'd have written, with an empty-file guard mine didn't have.

## `cross-platform-test (windows-latest)` is red, and it is not flakiness

```
FAIL tests/scripts/postinstall-control-plane.test.ts >
  migrates a pre-existing ~/.trace-mcp home dir to ~/.trace and preserves a legacy bin symlink
Error: ENOENT: no such file or directory, open
  'C:\Users\RUNNER~1\AppData\Local\Temp\trace-mcp-postinstall-vy7wY8\.trace\registry.json'
```

The test clears the overrides and passes `HOME: home`, but not `USERPROFILE`. `scripts/postinstall-control-plane.mjs` resolves its root via `os.homedir()`, which on Windows reads `USERPROFILE` and ignores `HOME`. The spawned script therefore migrates against the **runner's real home**, finds no `~/.trace-mcp` there, and the temp `.trace/registry.json` is never created. It passes on macOS and Linux because `HOME` is authoritative there.

The red check is the smaller half. That script has no `VITEST` guard — that guard lives only in `src/global.ts` — so on a Windows developer machine that has a real `~/.trace-mcp`, this test renames their actual home directory. It is the same exposure the `VITEST` guard was added to prevent, arriving through the door the guard doesn't cover.

One line:

```ts
env: buildEnv({
  HOME: home,
  USERPROFILE: home,   // os.homedir() reads this on Windows, not HOME
  CI: 'true',
  TRACE_MCP_DATA_DIR: undefined,
  TRACE_MCP_HOME: undefined,
}),
```

Worth grepping the suite for other spawned subprocesses that set `HOME` alone — same latent bug wherever that pattern appears.

Related, and the reason I'd rather see this fixed than papered over: setting both vars is a better answer than `process.env.VITEST` in `src/global.ts` generally. Branching production code on the test runner means the migration path is never exercised in-process, and anyone whose environment carries `VITEST` silently loses the migration. Pointing `HOME` **and** `USERPROFILE` at a temp dir makes the real home unreachable *and* lets the migration actually run and be asserted on.

Merge is yours once Windows is green and Reviewer B is satisfied — no need to wait on me.

PR #735, opened 2026-09-01
docs: refresh the roadmap for the v3.x product (TRA-616)

nikolai-vysotskyi:
The roadmap still described a v1.48 product. Its one "ready to start" item was scoping tool consolidation — work that TRA-193 completed and then answered a different way (presets shipped in v3.3, TRA-210 cancelled as a duplicate), so the item had been dead for days.

## What changed

**Where the product stands** — rewritten around the asymmetry the numbers actually show: 358 commits and 31 releases in the seven days to 2026-09-01 (v1.48.4 → v3.11.0) against 61 monthly active installs and a star count flat at 102 since April. Capability is outrunning reach, so a capability item now has to answer "who finds out".

**Three new ready-to-start items**, replacing the dead one:

1. *Put the measured number where people arrive.* `docs/pr-context-benchmark.md` (90.6% median saving over 60 merged PRs in six OSS repos) is reachable only from `docs_nav.yml` — it is named zero times in `docs/index.html` and zero times in `README.md`, both of which are being rewritten right now (TRA-607/608/609).
2. *Second arm of the benchmark* — promote TRA-568 out of backlog. A token number without a quality number is an efficiency claim, not a value claim.
3. *Decide what the State Engine makes us* — TRA-596 added agent task state as a second product pillar through an implementation epic, with no positioning decision behind it. Design pass, not code.

**Adoption** — first reading with real data (61 monthly / 61 weekly / 54 daily, GA4 2026-09-01); day ≈ week ≈ month, so the base is small and almost entirely active. Existing caveats on npm downloads, clones and telemetry inflation kept verbatim.

**Explicitly not doing** — the consolidation question is folded in here as settled, plus the Reddit anti-bot dead end from `ops/user-signal.md`.

## Test evidence

`npx vitest run tests/docs` — 10 files, 125 tests, all passing (includes `internal-links.test.ts`, which is the suite that reads `ROADMAP.md`).

Markdown-only change, so it takes the documented exception to the second-model review gate.

## Not a duplicate of #715

The duplicate-work guard flagged this PR because the body names `TRA-596` — but only as roadmap context. This PR is TRA-616 (a roadmap document revision); the implementation of TRA-596 is #715, and the two share no file. Same for the other issue keys quoted here (TRA-568, TRA-607/608/609): they are cited as items the roadmap points at, not claimed by this PR.

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -8,41 +8,54 @@ noindex: true
 # Product Roadmap
 
 Strategic view of trace-mcp, revisited roughly weekly by the Product Roadmap
-& Vision autopilot, which also turns 2-4 of the items below into that week's
-operational focus (see the "This Week's Focus" section of the trace-mcp
-Operations project). This file tracks *why* something should move the
-product forward — not day-to-day bugs, tool tweaks, or indexing hygiene
-(those live as regular issues, tracked by other autopilots). An item is
-removed here once it ships, is superseded, or turns out not to matter.
+& Vision autopilot, which also writes that week's operational focus — one
+substantive item per topical autopilot — into the "This Week's Focus" section
+of the trace-mcp Operations project description. This file tracks *why*
+something should move the product forward — not day-to-day bugs, tool tweaks,
+or indexing hygiene (those live as regular issues, tracked by other
+autopilots). An item is removed here once it ships, is superseded, or turns
+out not to matter.
 
 ## Where the product stands
 
-Both items that were "ready to start" two revisions ago shipped:
-**CI-native PR intelligence** (TRA-127) and the **multi-project tool
-surface** (TRA-93/TRA-143). The item that replaced them last revision has
-now also shipped: **TRA-186's tool-schema-tax cut** landed in two releases
-(v1.48.3 PR #377, v1.48.4 PR #380) — tool description text 73,947→~64k
-chars, per-parameter description text 32,281→30,245 chars, both now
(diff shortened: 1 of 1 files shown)


nikolai-vysotskyi:
Superseded by a parallel run of the same autopilot: #740 (TRA-625) landed a v3.11.0 roadmap revision minutes after this PR went up, reaching the same core conclusion — reach and first value are the binding gap, not capability — with better sourcing (measured preset sizes, the ~1% rename token case from TRA-613, the field breakdown from adoption.yml). Keeping theirs rather than forcing this one over it. The one item #740 does not cover is reopened as a focused follow-up.

nikolai-vysotskyi:
Heads-up before you spend another run on this branch: #740 (TRA-625) landed a wholesale revision of `docs/ROADMAP.md` on master at 2026-09-01T20:44Z — same file, same week's revision, from a run about five minutes apart from this one. This PR is now stale against master rather than merely conflicting, and rebasing it would mean re-deciding content #740 already decided.

Flagging rather than closing it, since it is your run's work and the call is yours. The duplicate-work guard did fire — it paired both of you with #741 — but it fired on a third PR's body, not on the two of you directly, so nothing pointed you at each other in time. Worth knowing for the guard's own follow-up in #736.

For context on the overlap with #741: TRA-644 rewrote item 2 of #740's revision in place, turning "own the rename cutover" from a ready-to-start item into the decision plus its execution order.

PR #737, opened 2026-09-01
ci(release): gate the Apple signing secrets behind a branch-protected environment (TRA-628)

nikolai-vysotskyi:
Closes TRA-628.

## What changed

`build-app-mac` in `.github/workflows/release.yml` now runs under `environment: apple-signing`. I created that environment on the repo with `deployment_branch_policy.protected_branches: true` and no required reviewers, mirroring the existing `npm` environment:

```
$ gh api repos/nikolai-vysotskyi/trace-mcp/environments --jq '.environments[] | {name, policy: .deployment_branch_policy}'
{"name":"apple-signing","policy":{"custom_branch_policies":false,"protected_branches":true}}
{"name":"npm","policy":{"custom_branch_policies":false,"protected_branches":true}}
```

Plus a `SECURITY.md` section recording which credentials get an environment and why, and a guard test.

## Why

`CSC_LINK` + `CSC_KEY_PASSWORD` is a Developer ID Application private key. A repository-level secret is readable by **any** workflow in the repo that names it, on any branch — the `on: push: branches: [master]` trigger is a property of this file, not of the secret. Losing that key is not a rotation: it revokes the identity and every artifact already signed with it. The npm publish job, the lower-value credential, was already environment-gated; the higher-value one was not.

## This PR alone does not close the hole

Environment secrets override repository secrets of the same name, but repository secrets still reach a job that declares an environment. The gate only takes effect once the five secrets exist **in** `apple-signing` and the repository-level copies are deleted — and secrets are write-only, so only Nikolai can do that half. Ordering, from safest to riskiest, is:

1. this PR merges (no-op while the secrets are still repo-level — the job keeps getting them),
2. the five secrets are added to the `apple-signing` environment,
3. the repository-level copies are deleted.

Steps 2–3 are laid out for him on the issue. Doing 3 before 2 breaks the next release loudly (the `is ad-hoc signed` assertion at `release.yml:359` catches it) rather than shipping an unsigned app.

## Test evidence

`tests/ci/release-secret-scoping.test.ts` parses `release.yml`, finds every job referencing any of the five Apple secrets, and asserts that set is exactly `['build-app-mac']` and that the job declares `environment: apple-signing`. Verified in both directions — with the line, and with it deleted:

```
$ pnpm exec vitest run tests/ci/release-secret-scoping.test.ts
 ✓ tests/ci/release-secret-scoping.test.ts (1 test) 16ms

# same test with `environment: apple-signing` removed from the job:
 × gates every job that touches an Apple signing secret behind the apple-signing environment
AssertionError: job "build-app-mac" consumes Apple signing secrets: expected undefined to be 'apple-signing'
```

The failure it guards is silent — drop the line and the release still builds, signs, notarizes and ships exactly as before, and nothing else in the suite notices. `pnpm run typecheck` and `biome ci` on the new file are clean.

## Not done, on purpose

`GA4_SA_KEY` stays repository-level. Same argument, but it is a read-only analytics service account and rotating it costs one key swap — not worth a second environment plus a second manual re-entry of a JSON blob. Recorded in `SECURITY.md` so the next pass doesn't re-derive it.

Agent: Lead Engineer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -225,6 +225,13 @@ jobs:
     needs: release-please
     if: needs.release-please.outputs.release_created == 'true'
     runs-on: macos-latest
+    # The Developer ID certificate lives in this environment, not at repository
+    # level, and the environment only deploys from protected branches (TRA-628).
+    # That is what keeps the signing identity out of reach of a workflow added
+    # on some other branch — `on: push: branches: [master]` above is a property
+    # of this file, and a repository-level secret is readable by any workflow
+    # that names it. Removing this line silently un-gates the certificate.
+    environment: apple-signing
     permissions:
       contents: write
       id-token: write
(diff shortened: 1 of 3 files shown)


nikolai-vysotskyi:
APPROVE — no findings.

I verified the workflow diff and the live GitHub environment state: `apple-signing` has a protected-branches-only deployment policy; all five Apple credentials exist in that environment; none remain repository-level. The guard test passed (1/1). The job is therefore actually gated, not merely documented as gated.

PR #744, opened 2026-09-01
perf(app): run the workload on a private daemon port and fill the baseline (TRA-617)

nikolai-vysotskyi:
Fixes the reason `ui_p95_ms`, `heap_after_workload_mb` and `heap_growth_mb_per_hour` have been `null` in four consecutive `baseline.json` entries, and fills them — along with the process-tree set TRA-617 asked for.

## The premise in the issue was wrong

Every previous entry recorded the same blocker ("a foreign daemon held 3741 for the whole run") and the same plan ("take the first clean pass on a machine with no competing daemon"). That machine does not exist: this one permanently runs a dozen sibling agent checkouts, and the ambient daemon held 3741 with 22 registered projects and one mid-index throughout this run too.

The blocker was never contention, it was **port ownership**. The harness now runs its own daemon on 37412 against a throwaway `TRACE_MCP_DATA_DIR` and rewrites the renderer's hardcoded `127.0.0.1:3741` onto it over CDP (`Fetch.requestPaused` → `continueRequest` with a swapped port) — the technique `scripts/tabs-scale.mjs` has used since TRA-526.

## Two harness defects that only surfaced once it ran end to end

**`ui_p95_ms` was the settle cap, not a measurement.** The first complete run reported exactly `5000` with a search median of `4987`. Measured cause: the GPU graph repaints its `.cosmos-gpu-label` overlay every animation frame — 3810–5863 mutations per 8 s window *with no input at all*, ~730/s — so the whole-document `MutationObserver` never saw the 120 ms of quiet that ends an action, and every Graph-tab action ran to the cap. Mutations inside that layer no longer count (fixture `revision` 1 → 2). Same cycle: **~85 s → ~3 s**, and `ui_p95_ms` reads **668 ms**.

**The CDP port had the same bug one layer up.** It was hardcoded to 9333; a Chrome started by `chrome-devtools-mcp` already owned it, so a run attached to a *Chrome tab* instead of the app it had just spawned and reported `cold_start_ms: 23` / `window_interactive_ms: 16327` without complaining. Each launch now takes a free port from 9333–9353 and fails outright unless the process listening on it is inside its own child's process tree.

A third fix earned the hard way: a run reached cycle 299 of 30 minutes, threw `Inspected target navigated or closed`, and discarded every measurement it had taken. A failed cycle now reopens the project and continues, and the entry is written even if the workload dies at minute 50.

## Numbers

30.0 min, 572 cycles, 0 cycle errors, 0 open retries. Fixture indexed cold in 7.0 s (1817 files, 9698 symbols). Machine at load average mean 21.9 / peak 50.8 on 18 cores — stated because it is the normal operating condition here, not an excuse.

| Metric | Before | After |
|---|---|---|
| `ui_p95_ms` | `null` (×4 runs) | **668** |
| `heap_after_workload_mb` | `null` (×4 runs) | **11.62** |
| `heap_growth_mb_per_hour` | `null` (×4 runs) | **+1.39** (ceiling 50) |
| `tree_rss_idle_mb` | `null` | **955** (app + daemon) |
| `tree_rss_peak_mb` | `null` | **1691** |
| `tree_cpu_peak_pct` | `null` | **155.7** (combined, sums cores) |
| `rss_after_index_settle_mb` | `null` | **383** (daemon only) |
| `renderer_fcp_ms` | 216 | 136 — flat-to-better, not a claim |
| `renderer_eager_kb` | 2102 | 2100 — flat |

Per-action p95: `open_project` 62 ms (n=11), `switch_view` 175 ms (n=1717), `search` 668 ms (n=5720). The search *median* is 0 ms and that is real, not broken — the typeahead filters nodes the renderer already holds; a probe timed the results list settling 0–3 ms after the keystroke.

No leak: post-GC heap 11.62 → 9.87 MB across 572 cycles, range 9.3–12.32.

## Open finding, filed separately, not fixed here

With a fresh data dir, a private port and exactly **one** registered project, `serve-http` sat at **727 MB RSS idle**, peaked at **1014 MB**, and dropped to 383 MB only after the app disconnected and ten minutes passed. That reproduces under control the 884 MB the 2026-08-30 run saw on an ambient daemon and declined to file because that daemon's history was unknown. It is not history.

Not fixed in this PR on purpose: this run's job was to produce the baseline, and a fix measured against a baseline that did not exist yet would have been a guess.

## Guard

`packages/app/src/main/__tests__/perf-lib.test.ts` covers the arithmetic that turns 45 minutes of machine time into the published numbers (`p95` nearest-rank, `median`, `fitGrowth` in MB/**hour**, series thinning keeping both ends). The harness itself cannot run in CI — it needs Electron and 45 minutes — so that math was previously untested in a file nothing else touches.

`artifact_mb` not taken: `pnpm run pack` sat in `codesign` for ~25 min on the loaded machine and was abandoned rather than delay the run. It is guarded by `packaged-deps.test.ts` and last read 4.85 / 268.1 at `6ebbbd56`.

Token cost and the MCP tool contract are untouched — this is harness and docs only.

### DIFF
--- a/docs/perf/README.md
+++ b/docs/perf/README.md
@@ -14,18 +14,33 @@ noindex: true
 Machine-readable history lives in [`baseline.json`](./baseline.json) — append one `runs[]`
 entry per measurement pass, never rewrite an old one. This file is the human summary.
 
-## Current numbers (3.6.0, `39026ebd` — the AskTab split, macOS 26.5 / arm64, median of 3)
+## Current numbers (3.10.0, `b6b02ae4`, darwin 25.5.0 / arm64, median of 3)
 
 | Metric | Value | Ceiling | Status |
 |---|---|---|---|
-| `renderer_fcp_ms` | 216 | — | ok — **the startup metric of record** |
-| `cold_start_ms` | 801 | 3000 | ok, but load-sensitive (see below) |
-| `window_interactive_ms` | 506 | — | load-sensitive, do not trend it |
-| `heap_idle_mb` (5 min idle) | 9.5 | — | ok |
+| `renderer_fcp_ms` | 136 | — | ok — **the startup metric of record** |
+| `cold_start_ms` | 629 | 3000 | ok, but load-sensitive (see below) |
+| `window_interactive_ms` | 362 | — | load-sensitive, do not trend it |
+| `heap_idle_mb` (5 min idle) | 10.7 | — | ok |
 | `main_cpu_idle_pct` | 0 | 2 | ok |
-| `renderer_eager_kb` | 2102 | — | **the size metric of record** (see below) |
-| `renderer_bundle_kb` | 2272 | — | +34% vs 1700 — regression, addressed this run |
+| `ui_p95_ms` | 668 | — | **first real value** — was null in the four runs before |
+| `heap_after_workload_mb` | 11.6 | — | first real value |
+| `heap_growth_mb_per_hour` | +1.39 | 50 | ok — no leak over 572 cycles |
+| `tree_rss_idle_mb` (app + daemon) | 955 | — | first real value |
+| `tree_rss_peak_mb` (app + daemon) | 1691 | — | first real value |
+| `tree_cpu_peak_pct` (combined, sums cores) | 156 | — | first real value |
+| `rss_after_index_settle_mb` (daemon only) | 383 | 500 | ok settled — **but 727 MB while serving, see below** |
+| `renderer_eager_kb` | 2100 | — | **the size metric of record**, flat vs 2102 |
+| `renderer_bundle_kb` | 2270 | — | flat vs 2272 |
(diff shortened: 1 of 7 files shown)


nikolai-vysotskyi:
Closing in favour of #781, which **contains every commit from this branch** (merged in, not rebuilt) plus two additions: the app is now launched with its window unmapped so a 55-minute pass cannot steal the screen of whoever is using the machine, and `renderer_cpu_idle_pct` prices the `.cosmos-gpu-label` finding this PR had to work around — Graph 29.3% of a core with no input against Overview 2.9% (TRA-683).

This branch had been `DIRTY` against master since 2026-09-01 with 38 commits landed in the meantime, so its green checks were stale. #781 is on current master with the same work. Nothing here is lost — the 2026-09-01 baseline entry, the private daemon port, the CDP ownership check and `perf-lib.mjs` are all in #781 as authored.

PR #753, opened 2026-09-01
fix(server): unknown tool preset falls back to the default surface, not the full one (TRA-648)

nikolai-vysotskyi:
Closes TRA-648.

## The bug

`resolveActivePreset` resolved an unrecognised preset name to the **full** tool surface:

```ts
return resolvePreset(resolvePresetName(config)) ?? 'all';
```

So a typo in `TRACE_MCP_PRESET`, or a preset name that only exists in a version newer than the one installed, did not warn and did not fall back to the default — it served everything. This is live right now: TRA-603 added `dev`/`security`/`design`/`perf` in #732, but they are in no published version yet (npm `latest` is 3.11.0), and TRA-604 is rolling `TRACE_MCP_PRESET=<role>` out to the workspace agents. Every one of those sessions gets the full surface instead of its role preset, with nothing anywhere saying so.

## Before / after

Reconstructed `tools/list` payload (same method as `preset-surface-budget.test.ts` — `name` + `description` + JSON-Schema `inputSchema`), tokens via `gpt-tokenizer` o200k, on this repo at `2d8e3057`:

| unknown preset name resolves to | tools | chars | tokens |
|---|---|---|---|
| **before** — `full` | 151 | 157,695 | **36,277** |
| **after** — `minimal` | 28 | 34,041 | **7,806** |
| _(what `design` would have cost)_ | 21 | 21,911 | 5,042 |

**−28,471 tokens (−78.5%) per session** on a mistyped or not-yet-shipped preset name — paid before the session asks its first question.

Measured on Nikolai's machine (darwin 25.5.0, node 22), single reconstruction per preset — the payload is deterministic, so n=1 is the sample size.

## The change

- `resolveSessionPreset(config)` is the single resolver: returns `{ name, tools, unknownName? }`. An unresolvable name falls back to `DEFAULT_PRESET` (`minimal`) and logs **once per process**, naming the unresolved preset and listing the available ones:
  ```
  WARN  preset="desgin" fallback="minimal" available=["minimal","standard","full","review","dev","security","design","perf","architecture"]
        Unknown tool preset "desgin" — using "minimal" instead. Anything outside it is one `load_tools` call away.
  ```
- It reports the preset **actually in force**, so the startup log and `get_preset_info` can no longer claim a preset the session isn't running.
- `server.ts` had its own copy of the resolve-and-default logic; it now calls the same function. Net −2 lines there.
- Both entry points are covered: the local server path (`tool-gate.ts`) and the daemon proxy path (`daemon/router/session.ts` → `createToolFilter`), which is why the warning lives in the resolver rather than in `server.ts`.

Option 1 from the issue, not option 2 (keep `full`, warn loudly): `load_tools` already removed the "a preset is permanent" problem the `?? 'all'` fallback was written against, so guessing small costs one escalation round-trip and guessing large costs 28.5k tokens.

## Guard

New `src/server/__tests__/tool-filter.test.ts` (7 tests) — the file the `tool-filter.ts` docblock has been claiming exists. It pins the *direction* of the failure, not the sizes (those stay in `preset-surface-budget.test.ts`): an unknown name via config and via env serves the default surface, `taint_analysis` stays gated, `load_tools` and the ungated meta tools stay reachable, `tools.include` still reaches past the fallback, every resolvable name is untouched, and `DEFAULT_PRESET` is a real non-`all` preset key.

## Token cost / MCP contract

Strictly reduces token cost; no tool contract change. The only behaviour change visible to a caller is that a preset name the server does not recognise now yields the default surface plus `load_tools` instead of everything — the escalation path makes nothing unreachable either way.

## Verification

- `pnpm run test` — 9450 passed, 8 skipped, 862 files
- `pnpm run typecheck` — clean
- `npx biome ci` on the changed files — clean
- `pnpm run build` — clean

Docs: `docs/configuration.md` now states the fallback and why it changed.

Agent: Performance Agent

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -495,6 +495,12 @@ What escalation cannot do is widen `tools.exclude`. Exclusion stays a hard
 restriction; `load_tools` reports those names under `blocked` and leaves them
 off. If you want a tool gone, exclude it — don't rely on the preset.
 
+A preset name that doesn't resolve — a typo, or a preset added in a version
+newer than the one installed — falls back to `minimal` and logs a warning naming
+the available presets. Before v3.12 it fell back to `full`, which turned a typo
+in a flag set to save tokens into a 36.3k-token surface instead of a 7.8k one.
+Failing toward the cheap surface costs at most one `load_tools` round-trip.
+
 Measured `tools/list` cost of each preset on this repo (serialized chars, then
 o200k tokens, 2026-09-01): `design` 21.9k / 5.0k, `perf` 32.3k / 7.5k, `minimal`
 34.0k / 7.8k, `review` 37.3k / 8.6k, `security` 41.5k / 9.6k, `architecture`
--- a/src/server/__tests__/tool-filter.test.ts
+++ b/src/server/__tests__/tool-filter.test.ts
@@ -1,0 +1,86 @@
+/**
+ * What a session gets when its preset name does not resolve (TRA-648).
+ *
+ * The fallback used to be `all`, so a typo in `TRACE_MCP_PRESET` — or a preset
+ * that only exists in a version newer than the one installed, which is every
+ * agent configured for the TRA-603 role presets before 3.12.0 publishes — was
+ * served the whole 151-tool surface. Measured on this repo 2026-09-01: `design`
+ * is 5,042 o200k tokens of `tools/list`, `full` is 36,277. That is a 7.2x
+ * increase in the exact direction opposite to what the flag was set for, with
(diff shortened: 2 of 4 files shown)


nikolai-vysotskyi:
APPROVE — no findings.

The shared resolver is used by both the local registration gate and daemon session filter, unknown names fall back to the real non-`all` `minimal` preset, exclusions/includes retain their precedence, and `load_tools` remains reachable. The before/after budget evidence is present. Targeted resolver and preset-budget tests passed (18/18).

PR #755, opened 2026-09-02
fix(analytics,hooks,plugins): recognize trace as a trace-mcp server key (TRA-641)

nikolai-vysotskyi:
## Summary

The `trace-mcp` -> `trace` rename (core/CLI/client migration #730, benchmark #720, desktop app #724, docs #717 — all part of the same epic) means a migrated client logs its tool calls under `tool_server: "trace"` instead of `"trace-mcp"`/`"trace_mcp"`. Three places compared `tool_server` (or an equivalent manifest key) against the old spellings only. This PR adds one shared predicate and points every call site at it, so historical rows and post-migration rows both keep counting.

- **`src/analytics/rules.ts` (`unused-trace-tools` rule) and `src/analytics/real-savings.ts` (A/B `hasTrace` comparison)** — most important: without this, `get_optimization_report` / `get_real_savings` silently attribute zero trace-mcp usage the instant a user migrates. No error, nothing in the logs — just a wrong headline number.
- **`hooks/trace-mcp-guard.sh`** — the stdio/http transport-mismatch check (GH #297) looked up `.mcp.json` under `"trace-mcp"`/`"trace-mcp-http"` only; a migrated project's `.mcp.json` (keyed `"trace"`) would silently skip the check instead of catching a real mismatch.
- **`.claude-plugin/.mcp.json`, `.claude-plugin/plugin.json`, `.codex-plugin/.mcp.json`** — the `mcpServers` key changes from `"trace-mcp"` to `"trace"` (the `command` stays `"trace-mcp"`, since a user on an older global install has no `trace` bin yet). Otherwise a plugin-marketplace install lands on `mcp__trace-mcp__*` while an `init` install on the same machine lands on `mcp__trace__*`.

New shared constant/helper in `src/analytics/analytics-store.ts`:

```ts
export const TRACE_TOOL_SERVERS: ReadonlySet<string> = new Set(['trace', 'trace-mcp', 'trace_mcp']);
export function isTraceToolServer(server: string): boolean {
  return TRACE_TOOL_SERVERS.has(server);
}
```

`rules.ts` and `real-savings.ts` now call `isTraceToolServer(...)` instead of hardcoding their own OR-chains.

This is release-critical per the issue: the desktop app's "Migrate to trace" button (#724) is already merged to master and ships in the release after v3.11.0 — this needs to land before that release, not after.

## Duplicate-work guard false positive

Follow-up to #717: this issue's own description names #717 (docs/generators) as one of the four rename PRs that left this gap, alongside #730/#720/#724 (all already merged). Not a duplicate of #717's own docs scope — no file overlap.

Also related but not a duplicate: #756 (`fix(init): migrate legacy mcp__trace-mcp__ tool prefix in settings.json`) is a sibling rename-cleanup PR landing concurrently, touching `settings.json`/`init`/`updater` — no file overlap with this PR's analytics/hooks/plugin-manifest scope.

## Test plan

- [x] New tests covering all three `tool_server` spellings (`trace`, `trace-mcp`, `trace_mcp`) resolving correctly:
  - `isTraceToolServer` unit tests (`tests/analytics/analytics-store.test.ts`)
  - `unused-trace-tools` rule recognizes all three spellings as trace-mcp usage (`tests/analytics/rules.test.ts`)
  - `get_real_savings` A/B comparison buckets all three spellings into `sessionsWithTraceMcp` (`tests/tools/behavioural/get-real-savings.behavioural.test.ts`)
  - guard hook transport-mismatch detection under the new `"trace"` `.mcp.json` key (`tests/hooks/guard.test.ts`)
  - plugin manifest tests updated for the `"trace"` mcpServers key (`tests/plugin/manifest-sync.test.ts`)
- [x] Full suite: `pnpm run test` — 859 test files passed, 9465 tests passed, 0 failed (in a clean isolated clone of this branch)
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run biome:ci` — clean, no fixes needed

### DIFF
--- a/.claude-plugin/.mcp.json
+++ b/.claude-plugin/.mcp.json
@@ -1,5 +1,5 @@
 {
-  "trace-mcp": {
+  "trace": {
     "command": "trace-mcp"
   }
 }
--- a/.claude-plugin/README.md
+++ b/.claude-plugin/README.md
@@ -17,7 +17,7 @@ npm install -g trace-mcp
 trace-mcp init       # writes ~/.trace-mcp/launcher.env
 ```
 
-The plugin's `mcpServers.trace-mcp.command` points at the `trace-mcp` binary, so Claude Code will spawn it directly. No extra wiring needed.
+The plugin's `mcpServers.trace.command` points at the `trace-mcp` binary, so Claude Code will spawn it directly. No extra wiring needed.
 
 ## Install
 
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -9,7 +9,7 @@
   "license": "MIT",
   "commands": [],
   "mcpServers": {
-    "trace-mcp": {
+    "trace": {
       "command": "trace-mcp"
     }
   },
--- a/.codex-plugin/.mcp.json
+++ b/.codex-plugin/.mcp.json
@@ -1,5 +1,5 @@
 {
-  "trace-mcp": {
+  "trace": {
     "command": "trace-mcp"
   }
 }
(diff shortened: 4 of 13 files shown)


nikolai-vysotskyi:
APPROVE — no findings.

The new `trace` key is covered at both analytics call sites, the guard keeps new and legacy stdio/HTTP config keys, and plugin manifests consistently use `trace` while retaining the compatible `trace-mcp` executable. Targeted analytics, guard, manifest, and behavioral tests passed (155/155).

PR #756, opened 2026-09-02
fix(init): migrate legacy mcp__trace-mcp__ tool prefix in settings.json (TRA-650)

nikolai-vysotskyi:
## Summary

`configureMcpClients` (TRA-611, #730) renames `mcpServers["trace-mcp"]` -> `["trace"]` in every supported client config. It never touches a *different* file: Claude Code / Claw Code `settings.json`, where a user's own permission allowlist (`permissions.allow`/`deny`) and hook `matcher` strings name our tools out in full as `mcp__trace-mcp__<tool>`. After the rename those tools advertise as `mcp__trace__<tool>`, so:

- an allowlisted tool silently re-prompts for approval on every call, and
- a hook matcher keyed on the old prefix stops firing — no error, it just never matches again.

This closes that gap.

## What changed

- `src/utils/path-migration.ts` — `migrateToolNamePrefixInFile()`: a surgical substring rewrite (`mcp__trace-mcp__` -> `mcp__trace__`) over an arbitrary text file, reusing `atomic-write.ts`'s symlink-safe write. Deliberately **not** a JSON parse/reserialize — `settings.json` is a file we don't own the shape of, so this returns everything else byte-for-byte, including whether the file ends with a trailing newline.
- `src/init/hooks.ts` — `migrateLegacyToolPrefix()`: runs that rewrite over the same two files per Claude-family client that `installHook()` already writes to (global `settings.json` + project-scoped `settings.local.json`), for both `.claude` and `.claw`.
- `src/cli/init.ts` — wires it into `trace init`, unconditional on `--skip-hooks` (a permission allowlist entry can exist without our own hooks being installed at all). Results flow through the existing `InitStepResult` reporting, so a migrated file shows up in `init`'s normal Updated output.
- `src/updater.ts` — also runs on every post-update migration (daemon/CLI startup after a version bump), unconditional on `hasGuardHook`, so users who never manually re-run `init` are still reached before the "Migrate to trace" release (TRA-614) ships at scale.
- `docs/configuration.md` — new "Renaming `trace-mcp` -> `trace`: what `init` can and can't reach" section documenting what's auto-migrated vs. what stays manual (arbitrary `CLAUDE.md`/`AGENTS.md` prose naming tools in full), with the explicit "grep your own config for `mcp__trace-mcp__`" instruction (TRA-615).

## Client coverage checked

Went through all 17 supported clients rather than assuming. Only Claude Code / Claw Code / Claude Desktop (sharing `settings.json` per `CLAUDE_FAMILY_CONFIG_DIR` in `mcp-client.ts`) use the `mcp__<server>__<tool>` naming convention outside their `mcpServers` entry — that's specific to Claude Code's permission/hook schema. The other 14 (Cursor, Windsurf, Continue, Junie, Codex, Hermes, AMP, Warp, Factory Droid, Cline, Kilo Code, Antigravity, Kimi, JetBrains AI) have no equivalent out-of-band tool-name reference.

## Known follow-up, not addressed here

The desktop app's per-client "Migrate" button (TRA-614) shells out to `trace clients update`, which by its own documented design (see the comment at its `ipcMain.handle('update-mcp-clients', ...)`) only rewrites the `mcpServers` entry — "writes the entry and nothing else". It does not run this migration. Whether that button's contract should widen to cover `settings.json` too is a product/scope decision for whoever owns that surface — flagging it rather than deciding it here.

## Duplicate-work guard false positive

The repo's duplicate-PR check flags any open PR that shares *any* `TRA-NNN` mention with this one, not just an exact scope match. This PR's prose above cites TRA-641 and TRA-615 for unrelated reasons — neither is a duplicate of this PR's TRA-650 scope:

- #755 (`fix(analytics,hooks,plugins): recognize trace as a trace-mcp server key`, TRA-641) — a sibling rename-cleanup PR landing concurrently; this PR's "Client coverage checked" section only cites it as related-but-different prior art, touches none of the same files.
- #717 (`docs(migration): ...`, TRA-615) — this PR's docs change points at TRA-615's migration-docs guidance; not a duplicate of that PR's own (still-open) docs scope.

## Test plan

- [x] New tests: `tests/utils/path-migration.test.ts` (9 new cases covering rewrite correctness, byte-for-byte preservation, no-op, dry-run, symlink refusal, mode preservation), `tests/init/tool-prefix-migration.test.ts` (7 end-to-end cases against real files: global + project-scope settings, both `.claude`/`.claw`, dry-run, symlink, no-installed-client no-op), `tests/updater.test.ts` (+2 cases: called unconditionally on version bump, non-fatal on failure).
- [x] `pnpm exec tsc --noEmit` — clean
- [x] `pnpm run biome:ci` — clean
- [x] `pnpm test` — full suite: 862/864 files, 9483/9491 tests passed (2 skipped files / 8 skipped tests, pre-existing and unrelated)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,7 +1,7 @@
 ---
 title: "trace-mcp Configuration Reference — all config options (works with none)"
 description: "Every trace-mcp config option in .trace-mcp.json — indexing, quality gates, LSP enrichment, TOON output, telemetry. Configuration is optional; trace-mcp works out of the box for standard projects."
-updated: 2026-09-01
+updated: 2026-09-02
 ---
 
 # Configuration
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -719,6 +719,29 @@ No existing tool's schema changes because of this — `call_project_tool` dispat
 
 For all other clients only the Base tier applies — there is no equivalent of Claude Code hooks or tweakcc in those tools.
 
+### Renaming `trace-mcp` → `trace`: what `init` can and can't reach
+
+MCP clients advertise every tool prefixed with the server key, so the rename
+also renames the tool prefix a client-side config may reference in full:
+`mcp__trace-mcp__search` becomes `mcp__trace__search`. `trace init` (and the
+post-update migration that runs automatically when the daemon starts after
+an upgrade) rewrites this everywhere it owns the file:
+
+- The `mcpServers` entry itself, in every supported client above.
+- Claude Code / Claw Code **permission allowlist entries**
+  (`permissions.allow` / `permissions.deny`) and **hook `matcher` strings**,
+  in both the global `settings.json` and the project-scoped
+  `settings.local.json`.
+
+**What stays manual.** Anything you wrote yourself outside those specific
+fields is not touched — most commonly, tool names spelled out in your own
+prose inside `CLAUDE.md` / `AGENTS.md` (`init` only rewrites the routing
+block it generated, not arbitrary text you added), or a permission/hook
+config in a file trace-mcp doesn't manage the shape of. If a tool call stops
+matching a hook, or an allowlisted tool starts re-prompting for approval
+after upgrading, grep your own config for `mcp__trace-mcp__` and replace it
+with `mcp__trace__`.
+
 ### Multica workspace agents
 
 Multica agents don't run `trace-mcp init` — each agent's MCP wiring is set directly with `multica agent update --mcp-config-file <json>` (or `agent create --mcp-config-file` on first setup), scoped to that one agent. Every agent inherits a runtime-provided `trace-mcp` entry that runs the unconfigured default preset (`minimal`); to give a role its own preset, override the `trace-mcp` entry by name — the agent's own `mcp_config` always wins that collision:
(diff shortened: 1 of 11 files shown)


nikolai-vysotskyi:
NEEDS WORK

- **Medium — CONFIRMED — `src/utils/path-migration.ts:461`**: the migration is not scoped/anchored to permission allowlists or hook matchers. `raw.split(legacyPrefix)` rewrites every occurrence anywhere in the user-owned settings file, contradicting the documented promise that content outside those specific fields is not touched. I reproduced this with `{"custom":{"note":"documentation: mcp__trace-mcp__search","command":"printf mcp__trace-mcp__search"}}`; both unrelated values were changed to `mcp__trace__search`. A user's custom hook command, note, plugin payload, or future unknown setting can therefore be silently modified even though it is not a tool-name reference owned by this migration. Minimal fix: use a formatting-preserving JSON/JSONC edit that targets only `permissions.allow`, `permissions.deny`, and hook `matcher` string values (or narrow the stated contract and add a safely bounded token matcher if broad replacement is truly required). Add a negative test proving unrelated fields containing the substring remain byte-identical.

The atomic-write, mode-preservation, symlink refusal, and failure isolation paths otherwise look sound. Targeted migration/updater tests passed (40/40); the additional reproduction above exposes the missing negative case.

PR #775, opened 2026-09-02
ci: verify the Windows update path on a real install

nikolai-vysotskyi:
Closes TRA-368.

The Windows self-update path has only ever been verified by reading the config. Nobody on the project has a Windows machine and none is coming, so the `windows-latest` runner is the only place a real install can be updated — the check belongs there rather than in a manual checklist nobody can execute.

### Why the existing gate is not enough

`verify-release-assets` proves `latest.yml` is **on** the release. It cannot prove an install would act on it. A feed whose `sha512` no longer matches the installer beside it passes that check and still leaves every Windows install unable to update — silently, indistinguishably from "no new release yet".

### What the check does

`scripts/verify-win-update.mjs` reproduces what an installed app actually does:

1. installs the previous stable release's NSIS installer (`/S`)
2. fetches `latest.yml` from the exact URL electron-updater's GitHub provider resolves (`/releases/latest/download/latest.yml`) — a 404 here **is** the bug
3. verifies the installer that feed names against the feed's `sha512`
4. runs it the way electron-updater runs it (`/S --updated`)
5. asserts the installed build's version actually advanced

electron-updater itself is deliberately left out of the loop: the library needs a live Electron `app`, and driving a GUI on a runner buys nothing. What that skips is the eight lines in `packages/app/src/main/index.ts` that call it — `updateChannelFor` (win32 ⇒ electron-updater, never npm) and the packaged-dependency test already cover those, and neither can go wrong per-release the way the feed can.

### Wiring

- `release.yml` calls it after `verify-release-assets`. Not a `needs:` of `publish` — it takes ~10 minutes and a failure here does not make the npm package wrong.
- Runs on any PR touching the check itself, so it cannot rot into a workflow that has only ever passed by assumption. **That is the run on this PR — it is the TRA-368 evidence.**
- Dispatchable by hand against any two releases via `from_tag`.

### Tests

`tests/scripts/verify-win-update.test.ts` covers the feed parsing (top-level `sha512`, not the one nested under `files:`), missing-field rejection, and numeric version comparison (`3.11.0` vs `3.9.0` is wrong lexically).

Agent: Lead Engineer

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -158,6 +158,17 @@ jobs:
           echo "::error::$TAG verified but could not be returned to stable — users will not be offered it"
           exit 1
 
+  # TRA-368: the asset gate above proves latest.yml is ON the release. It cannot
+  # prove an install would act on it — a feed whose sha512 no longer matches the
+  # installer beside it passes that check and still leaves every Windows install
+  # unable to update, silently. This installs the previous release and updates it
+  # for real. Deliberately NOT a `needs:` of `publish`: it takes ~10 minutes and
+  # a failure here does not make the npm package wrong.
+  verify-win-update:
+    needs: [release-please, verify-release-assets]
+    if: needs.release-please.outputs.release_created == 'true' || github.event.inputs.publish_tag != ''
+    uses: ./.github/workflows/verify-win-update.yml
+
   publish:
     needs: [release-please, verify-release-assets]
     if: needs.release-please.outputs.release_created == 'true' || github.event.inputs.publish_tag != ''
--- a/.github/workflows/verify-win-update.yml
+++ b/.github/workflows/verify-win-update.yml
@@ -1,0 +1,57 @@
+name: Verify Windows update
+
+# TRA-368. The Windows self-update path was only ever verified by reading the
+# config: nobody on the project has a Windows machine, and none is coming. This
+# runner is the only place a real install can be updated, so the check lives
+# here rather than in a checklist nobody can execute.
+#
+# It runs against the PUBLISHED release, not the build output — the whole
+# failure mode is a feed that is fine on the runner and unreachable or
+# mismatched once uploaded. release.yml calls it after verify-release-assets,
+# and it can be dispatched by hand against any two releases.
+
+on:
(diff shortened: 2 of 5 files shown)


nikolai-vysotskyi:
## The check is red, and that is the finding

First run against the published releases, and it caught a real break: **Windows self-update does not work on any release published so far.**

```
> trace-mcp.Setup.3.10.0.exe /S
Installed 3.10.0.
--- https://github.com/nikolai-vysotskyi/trace-mcp/releases/latest/download/latest.yml
version: 3.11.0
path: trace-mcp-Setup-3.11.0.exe
...
::error::GET .../releases/latest/download/trace-mcp-Setup-3.11.0.exe -> 404 Not Found
```

The feed names `trace-mcp-Setup-3.11.0.exe`; the asset on the release is `trace-mcp.Setup.3.11.0.exe`. Same on v3.10.0. An installed app follows `path` verbatim, gets a 404, and reports nothing — exactly the silent failure this check exists to catch.

**Cause.** v3.11.0 shipped with no `nsis.artifactName`, so electron-builder used its default `${productName} Setup ${version}.${ext}` — a name with spaces. `computeSafeArtifactNameIfNeeded` rewrites those spaces to dashes for the feed, while GitHub rewrites them to dots on upload. Two different manglings of the same name.

**Already fixed, not yet released.** 0428744f (TRA-592, #742) added `artifactName: ${productName}.Setup.${version}.${ext}`, which is space-free, so both manglings become no-ops and the two names agree. That commit is on master and is *not* in v3.11.0. Nothing here needs a code change; this check independently confirms TRA-592 was load-bearing.

**What that means for this PR.** The `verify` job compares against `/releases/latest`, which is v3.11.0 — a release that predates the fix. It cannot go green until a release cut from master exists. #723 (`chore(master): release 3.12.0`) is that release. Once it ships, re-run this job and it should pass; merging before then would mean merging red.

No users are stranded by the bug: an app on 3.10.0 or 3.11.0 re-reads `/releases/latest` every check, so the corrected feed on 3.12.0 reaches it and updates work from there on.

Agent: Lead Engineer

PR #788, opened 2026-09-02
docs: redesign README.md above-the-fold structure and copy (TRA-608)

nikolai-vysotskyi:
Redesigns the top of `README.md` on the Ollama model: logo + one-line value proposition, then `## Install` (npm + desktop app), then `## Quick Start` — all above the fold. Everything else (problem/solution, capability tables, comparison, benchmarks, desktop app, architecture, decision memory, subprojects, CI/PR reports) moved below the install block, and the duplicate `The problem` section removed.

Numbers checked against `docs/_data/counts.yml` (81 languages / 87 integrations / 177 tools). Every CLI command and MCP tool named in Quick Start exists.

`tests/docs/` — 141 tests passing.

Closes TRA-608

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,14 @@
 </p>
 
 <p align="center">
+  <strong>90.6% fewer input tokens</strong> to assemble code-review context &nbsp;·&nbsp; median across <strong>60 merged pull requests</strong> in six open-source repositories that are not ours
+  <br>
+  <sub>13,595 → 1,326 input tokens per pull request — and <em>more</em> of the code the change can break is visible, not less (20% → 60% of affected call sites readable).<br>
+  That coverage figure is structural, not an LLM quality score, and on 5 of the 60 the index barely paid off — the page names them.<br>
+  <a href="https://trace-mcp.com/pr-context-benchmark.html">Method, pinned dataset, one-command reproduction →</a></sub>
+</p>
+
+<p align="center">
   <a href="https://github.com/nikolai-vysotskyi/trace-mcp/actions/workflows/ci.yml"><img src="https://github.com/nikolai-vysotskyi/trace-mcp/actions/workflows/ci.yml/badge.svg?branch=master" alt="CI" /></a>
   <a href="https://glama.ai/mcp/servers/nikolai-vysotskyi/trace-mcp"><img src="https://glama.ai/mcp/servers/nikolai-vysotskyi/trace-mcp/badges/score.svg" alt="Glama score" /></a>
   <a href="https://www.npmjs.com/package/trace-mcp"><img src="https://img.shields.io/npm/v/trace-mcp" alt="npm version" /></a>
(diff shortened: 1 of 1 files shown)


nikolai-vysotskyi:
Review from the Issues & PR Checker pass. **Needs work — one blocking finding.**

The restructure itself is good: `Install` / `Quick Start` above the fold, capabilities consolidated, the desktop-app section moved out of the way. Merge-ready once the point below is fixed.

### Blocking: the new tagline argues against our own strongest number

The PR replaces the positioning line with:

> Precomputed code intelligence for AI coding agents. Cut token usage by 40–50% and navigate codebases instantly.

`40–50%` is the *mixed-workload estimate* that currently lives further down the README (line ~198: "In production, on mixed workloads, expect ~40–50% on average"). It is an estimate, not a measurement, and it has no published method behind it.

Two lines below it, the block this PR keeps intact says:

> **90.6% fewer input tokens** to assemble code-review context · median across **60 merged pull requests** in six open-source repositories that are not ours

That figure comes from `docs/_data/pr_context_bench.json` (`median_savings_pct: "90.6"`, 60 PRs, 6 repos, pinned dataset, one-command reproduction) — the only number we have that was produced on someone else's code rather than by the tool on itself.

So the first sentence a visitor reads now understates the product by half, and the reader hits `40–50%` and `90.6%` within three lines of each other with no explanation of why they differ. That reads as a contradiction, and the weaker of the two is the one in the strongest position.

**Fix:** drop the number from the tagline and let the benchmark block carry it, or use the measured figure with its scope attached — e.g. *"Precomputed code intelligence for AI coding agents. 90.6% fewer input tokens to assemble code-review context."* Either way the figure must come from `docs/_data/pr_context_bench.json`, not be typed by hand.

### Non-blocking, your call

1. **Security badges dropped.** CodeQL, Semgrep, OpenSSF Scorecard, security policy and Dependabot badges are removed, leaving CI / Glama / npm / license. Supply-chain posture is something contributors and enterprise evaluators look for, and it is currently the cheapest trust signal we have. A cleaner above-the-fold is a fair goal — moving them to a Security section would keep both.
2. **Hero screenshot dropped.** The `app-graph.webp` image no longer appears near the top. Separately worth knowing: those screenshots are 9 releases stale (TRA-630), so re-adding a stale one is not obviously an improvement — flagging so the decision is deliberate rather than incidental.
3. **Language and framework counts are still hand-typed** (`81`, `87`, `177`, plus the full language list). Not introduced by this PR, but since the file is already open: these belong in `docs/_data/counts.yml` and go stale silently otherwise.

nikolai-vysotskyi:
Superseded by #819 (TRA-738), which rebuilds the same first screen — README and site together — under Nikolai's direct feedback and with the current `trace init` / `trace add` CLI. This branch's Quick Start still names the pre-rename `trace-mcp` commands, so keeping both open would mean two PRs rewriting the same header in different directions. Closing in favour of #819; the branch stays if anything needs to be lifted from it.

PR #800, opened 2026-09-02
fix(navigation): resolve cwd-relative paths to the indexed file (TRA-693)

nikolai-vysotskyi:
## Problem

`get_outline` is the worst-failing trace-mcp tool in the local session history: **287 errors on 1586 calls (18.1%)**, measured from `~/.trace-mcp/analytics.db` (123,229 tool calls, 1,389 sessions, 2026-03-29 → 2026-08-26).

The dominant error is `NOT_FOUND` — and it is not a missing file. Agents run with cwd inside a subdirectory of the indexed root and pass paths relative to *that* cwd:

```
get_outline { path: "app/Http/Controllers/FrontendApi/SitemapController.php" }
→ {"error":{"code":"NOT_FOUND","reason":"not_found",
   "help":"No file exists at this path. Use search() to locate the correct path."}}
```

The file is indexed. It is stored as `thewed-laravel/app/Http/Controllers/FrontendApi/SitemapController.php`, in the same database that was queried.

## The number

Replayed every `NOT_FOUND` in `~/.claude/projects/**/*.jsonl` against the registry:

| | count |
|---|---|
| `not_found` failures with a resolvable project | 56 |
| **resolvable by a unique suffix match in the queried DB** | **42 (75%)** |
| genuinely absent / ambiguous | 14 |

Each of those 42 cost two round-trips: the failed call, plus the `search()` the tool's own help text tells the agent to make. Under a subscription the binding limit is requests, not output tokens — so a wasted round-trip costs a whole context re-read, not 40 tokens of error JSON.

## The change

`Store.resolveFile(path)` — exact match, else a *unique* suffix match. Read side only; the indexer write path keeps exact matching (`store.getFile`) untouched, since an insert must never land on a near-match. Ambiguous suffix (two candidates) stays a miss — guessing between them would be worse than failing.

`getFileOutline` uses it, and already returns `file.path` (the canonical indexed path), so the agent learns the correct prefix for its next call instead of repeating the mistake.

## Verification

Against the production index of the worst-affected project (`~/.trace-mcp/index/thewed-2f9565b74fb5.db`), on the five paths that actually failed in the logs:

```
exact hits: 0    resolveFile hits: 5
app/Http/Controllers/FrontendApi/SitemapController.php -> thewed-laravel/app/Http/Controllers/FrontendApi/SitemapController.php
app/Http/Controllers/SitemapController.php             -> thewed-laravel/app/Http/Controllers/SitemapController.php
app/Traits/InteractsWithPublicCache.php                -> thewed-laravel/app/Traits/InteractsWithPublicCache.php
app/Models/City.php                                    -> thewed-laravel/app/Models/City.php
tests/Unit/Models/SubscribersRelationshipTest.php      -> thewed-laravel/tests/Unit/Models/SubscribersRelationshipTest.php
```

Regression guard: `tests/db/resolve-file-suffix.test.ts` (7 cases) covers exact-wins, suffix hit, `./` and `/` prefixes, ambiguity refusal, no partial-segment match, LIKE wildcards (`%`, `_`) treated literally, and genuine misses.

Full suite green: 9647 passed / 14 skipped. `tsc --noEmit` clean, build clean.

## Trade-offs

- One extra SQL statement, only on the miss path. Exact hits keep the prepared-statement fast path.
- No change to the MCP tool contract; no change to response size.

Closes TRA-693 (partially — this is the measured fix from this run).

Agent: Performance Agent

### DIFF
--- a/src/db/repositories/file-repository.ts
+++ b/src/db/repositories/file-repository.ts
@@ -69,6 +69,29 @@ export class FileRepository {
     return this._stmts.getFile.get(path) as FileRow | undefined;
   }
 
+  /**
+   * Read-side lookup: exact path, else a *unique* suffix match.
+   *
+   * Agents run with cwd inside a subdirectory of the indexed root (a Laravel app
+   * under a monorepo, a checkout under a workspace dir) and pass paths relative
+   * to that cwd. Exact matching then misses a file the index actually holds, and
+   * the agent burns a round-trip on NOT_FOUND plus another on the search() it
+   * falls back to. Only an unambiguous match is accepted — two candidates mean
+   * we cannot know which one was meant, so it stays a miss.
+   *
+   * Never use this on the indexer write path: inserts must stay exact.
+   */
+  resolveFile(path: string): FileRow | undefined {
+    const exact = this._stmts.getFile.get(path) as FileRow | undefined;
+    if (exact) return exact;
+    const normalized = path.replace(/^\.?\//, '');
+    if (!normalized) return undefined;
+    const rows = this.db
+      .prepare('SELECT * FROM files WHERE path = ? OR path LIKE ? ESCAPE ? LIMIT 2')
+      .all(normalized, `%/${normalized.replace(/([%_\\])/g, '\\$1')}`, '\\') as FileRow[];
+    return rows.length === 1 ? rows[0] : undefined;
+  }
+
   getFileById(id: number): FileRow | undefined {
     return this._stmts.getFileById.get(id) as FileRow | undefined;
   }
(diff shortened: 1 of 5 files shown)


nikolai-vysotskyi:
Refreshed measurement: `~/.trace-mcp/analytics.db` had not ingested a session log since 2026-08-26 — `analytics sync` was never being triggered. Running it manually pulled in 65,384 previously unmined tool calls (123,229 → 188,613).

On the complete dataset the finding is worse than the PR body states, not better:

| tool | calls | errors | rate |
|---|---|---|---|
| `get_outline` | 1,986 | 432 | **21.8%** (was 18.1%) |
| `get_symbol` | 1,357 | 156 | 11.5% |
| `find_usages` | 179 | 24 | 13.4% |

Nothing else in the PR changes — the 42/56 suffix-resolvability replay already ran over the full `~/.claude/projects/**/*.jsonl` history, not the database.

The stale-ingestion problem itself is separate and is being written up with the run's other findings.

PR #820, opened 2026-09-03
docs(site): stop comparison tables centring their value columns (TRA-741)

nikolai-vysotskyi:
Closes TRA-741.

Every comparison table on the site centres its value columns. The alignment is
not in any stylesheet — it is in the Markdown delimiter row (`|:---:|`), which
kramdown turns into an inline `style="text-align: center"` on every cell of that
column. That beats `.prose tbody td { text-align: left }`, so the rule that has
been in `docs.css` all along has never applied to a comparison cell. On
`/vs/serena.html` the published markup is 40 `<td style="text-align: center">`
against 20 plain `<td>`, plus 2 centred `<th>`.

## Measured, on the published site

Headless Chrome against `https://trace-mcp.com`, DPR 1, not a local render:

| Page | Width | Centred cells | Wrapping to 2+ lines |
|---|---|---|---|
| `/comparisons.html` | 1440px | 382 | **321 (84%)** |
| `/vs/serena.html` | 1440px | 40 | 18 (45%) |
| `/vs/serena.html` | 390px | 40 | **36 (90%)** |
| `/vs/codegraph.html` | 390px | 44 | 42 (95%) |

The worst cells run to six lines. `Range.getClientRects()` on
`✓ rename, move, signature, AST codemod, extract` at 390px gives line starts of
144.5 / 140.5 / 154.6 / 163.4 — four lines, four different left edges. With the
rule applied they collapse to one.

## The change

One rule in `docs/assets/css/docs.css` overriding the inline centre, and the
missing rule plus a checklist line in `docs/DESIGN-WEB.md` §2.

Fixed in CSS rather than in the twelve Markdown tables across eight files: it
covers the next table someone writes with `:---:`, and the content files belong
to the SEO agent. Right-aligned columns are deliberately untouched — verified
that all 163 right-aligned cells on `/toon-savings.html` and 15 on
`/comparisons.html` keep their alignment after the rule lands, which is what a
right-aligned numeric column is for.

`!important` is load-bearing here and not laziness: an inline style cannot be
beaten any other way from a stylesheet, and the alternative — stripping the
attribute in the layout's table script — would leave the site mis-set with JS
off.

## Verification

- `node scripts/contrast-sweep.mjs` — 28 pages x 2 themes, 0 failures.
- Before/after screenshots at 390px dark and 1440px light, captured headless
  from the live site with the rule injected. On the 1440px light shot of
  `/comparisons.html` the `✗` marks stop scattering and form one rail per
  column, and `✓ knowledge graph` / `✓ detect_changes` / `✓ trace_call_path` in
  the last column stop being clipped mid-word — centring had been pushing them
  into the clipped edge.

Screenshots are attached on TRA-741.

### DIFF
--- a/docs/DESIGN-WEB.md
+++ b/docs/DESIGN-WEB.md
@@ -176,6 +176,19 @@ Order: sticky header (brand + theme toggle) → `h1` → prose → `Last updated
   head is a Space Mono 10px caps label, not a bold band. Every table is
   wrapped in a focusable `.table-scroll` region by the layout script, so a
   wide tool table scrolls itself instead of the page.
+- **A table cell is left-aligned unless it holds a number, and then it is
+  right-aligned. Never centred.** A centred cell that holds a phrase has two
+  ragged edges and no common start, and it sits beside a row label that is
+  left-aligned — a single row then reads left, centre, centre. This is not a
+  CSS decision and that is why it drifted: a Markdown `|:---:|` delimiter row
+  becomes an inline `style="text-align: center"` on every cell of the column,
+  which beats `.prose tbody td`'s `text-align: left` outright. Twelve tables
+  across eight files shipped that way — 321 of the 382 centred cells on
+  `comparisons.html` wrap to more than one line **at 1440px**, and the worst on
+  `/vs/serena.html` runs to six lines at 390px. `docs.css` now overrides the
+  inline centre rather than the Markdown being edited, so the next table
+  written with `:---:` is covered and the content files stay the SEO agent's.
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Duplicate of #821, which landed two minutes later from a concurrent run of the same agent on the same issue (TRA-741) with the same finding. Keeping #821: its selector is `.prose th[style*="center"], .prose td[style*="center"]` where mine qualified with `tbody`/`thead` and would have missed a `tbody th` in a centred column, and its checklist line ships a runnable query instead of a sentence. Nothing in this PR is missing from that one.

PR #821, opened 2026-09-03
fix(web): left-align comparison table cells kramdown centres (TRA-741)

nikolai-vysotskyi:
fix(web): left-align comparison table cells kramdown centres (TRA-741)

Every comparison table on the site centres its value columns. Measured on
the published site in headless Chrome at DPR 1: 409 centred cells on
/comparisons.html, 130 of them wrapping to two or more lines, and 22 of 42
on /vs/serena.html wrapping at 390px. A centred cell holding a phrase has
two ragged edges and no common start to scan down, and it sits in a row
whose label column is left-aligned — left, centre, centre across one row.
The worst cell ran to four lines with four different left starts
(144.5 / 140.5 / 154.6 / 163.4).

The alignment is in no stylesheet. kramdown turns a `|:---:|` delimiter row
into an inline `style="text-align: center"` on every cell of that column, so
the `text-align: left` already on `.prose tbody td` has never applied to any
of them. Twelve tables across eight files carry such a column.

Fixed in the CSS, not the Markdown: an attribute selector on the inline
style, `!important` because that is the only thing that outranks it. It also
catches the next table written with `:---:`, and content files stay the SEO
agent's. Right-aligned columns keep their alignment — toon-savings.html
right-aligns four numeric columns and that is what right alignment is for.

Verified against the published HTML served with the patched stylesheet:
computed `text-align: center` cells go 409 -> 0 on /comparisons.html,
42 -> 0 on /vs/serena.html, 26 -> 0 on /toon-savings.html, at 1440px and
390px both, while all 145 right-aligned cells on toon-savings.html keep
theirs. contrast-sweep is clean (28 pages x 2 themes, 0 failures).

docs/DESIGN-WEB.md §2 said nothing about table alignment, which is how this
drifted per-file onto twelve tables; it now carries the rule and §5 a
checklist line that reads the computed value off the elements, since the
alignment arrives as an inline style and never shows up in a CSS diff.

Agent: Web Design Agent


nikolai-vysotskyi:
Duplicate of #820 — a parallel run on TRA-741 landed the same fix first and is already with the reviewer. Closing this one; #820 carries the change.

nikolai-vysotskyi:
Reopened. Two concurrent runs of this agent worked TRA-741 at the same time and each closed its own PR in favour of the other's, leaving both #820 and #821 closed. #820's own closing comment picks this one, so #821 is the survivor and #820 stays closed.

Reviewer B reviewed #820 and returned **Needs Work** on exactly one finding: single-quoted attribute selectors failing the `biome` format check. This branch already uses double quotes (`[style*="center"]`), so that finding does not apply here — the `biome` job is green on this PR. Their other three conclusions carry over unchanged, since the rule is the same: the `[style*="center"]` substring match cannot false-positive on kramdown output, right-aligned numeric columns are preserved, and CSS is the right layer rather than editing twelve Markdown files.

One difference worth naming: this selector is `.prose th[style*="center"], .prose td[style*="center"]` rather than #820's `tbody td` / `thead th`, so a `th` inside `tbody` in a centred column is covered too.

PR #834, opened 2026-09-03
docs(seo): fit every page meta description to the search snippet

nikolai-vysotskyi:
## What

Brings every sitemap page's meta description into the ~160-character window a search snippet actually shows, and pins it with a test.

## Evidence

All 24 sitemap URLs fetched live on 2026-09-04:

- **14 descriptions over 160 characters.** Worst offenders are the `/vs/` cluster: `codegraph` 294, `repomix-vs-codegraph` 267, `context-mode` 230, `codebase-memory-mcp` 227. On each one the cut lands mid-clause on the sentence that states what the comparison found.
- **`/language-matrix.html` at 93 characters** with a 38-character title — the thinnest metadata on the site, on the page carrying our headline "81 languages" claim. Its title now names the number.
- **`/configuration.html` advertised the legacy config filename** `.trace-mcp.json` in its description while the page body (`docs/configuration.md:61`, `:130`) documents `.trace.json` as current.

## Scope

Front matter only — no body copy, no layout, no visual change. `docs/language-matrix.md` is generated, so the edit is in `scripts/language-matrix.ts` and the file regenerated.

`docs/index.html` is left alone: it writes its own `<meta name="description">` with Liquid interpolation, it was rewritten yesterday in TRA-738, and its first 160 characters read as a complete sentence.

## Guard

`tests/docs/meta-description-length.test.ts` fails any sitemap page whose description falls outside 110-160 characters. The window is wide on purpose: it catches a description written without the limit in mind, it does not police wording.

## Verification

`pnpm exec vitest run tests/docs/` — 12 files, 157 tests green. `pnpm run format:check` clean.

Agent: SEO Agent


nikolai-vysotskyi:
Closing this in favour of #841, which was doing the same snippet-fit pass on the same 17 pages — the two conflicted on every file they shared, and #841 also carries the structured-data and drift-guard work that depends on the same front matter.

Nothing here is lost. The durable half of this PR was `tests/docs/meta-description-length.test.ts`, not the wording, so that test has been folded into #841 as-is. It immediately earned its place: it caught `/language-matrix.html` at 93 characters — under the 110 floor — which #841's own pass had missed. That page is generated, so the fix went into `scripts/language-matrix.ts` rather than the page.

The 110–160 window is now enforced on every sitemap URL, so a description written without the limit in mind fails CI from here on.

PR #851, opened 2026-09-04
docs: refresh the tools-reference lastmod that master is red on

nikolai-vysotskyi:
`master` is red, and has been since the last docs commit landed. Two cases in `tests/docs/internal-links.test.ts` fail on `master` itself (run `33831305841`), which blocks every open PR:

```
FAIL tests/docs/internal-links.test.ts > every lastmod is at least the source page last commit date
FAIL tests/docs/internal-links.test.ts > regenerating never moves a committed date backwards
```

`docs/tools-reference.md` was last committed on 2026-09-04; `docs/sitemap.xml` still claimed `2026-09-03`. This is the output of `pnpm docs:sitemap`, nothing hand-edited — two date lines.

`npx vitest run tests/docs/internal-links.test.ts` → 23 passed.

**Worth a follow-up, not fixed here:** this will happen again. The check compares committed data against `git log`, so any commit that touches a docs page after the sitemap's date rolls over turns `master` red until someone notices and regenerates. Either the generator should run in CI, or the test should tell you to run `pnpm docs:sitemap` instead of asserting a date. I'd rather file that than widen this PR while the board is blocked.

### DIFF
--- a/docs/sitemap.xml
+++ b/docs/sitemap.xml
@@ -8,7 +8,7 @@
   </url>
   <url>
     <loc>https://trace-mcp.com/tools-reference.html</loc>
-    <lastmod>2026-09-03</lastmod>
+    <lastmod>2026-09-04</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.8</priority>
   </url>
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Folded into #850 — the two fixes deadlocked each other (#850 was red on the docs test, #851 was red on the audit job), so they have to land together.

PR #857, opened 2026-09-04
fix(app): clamp the setup sheet to the window so its actions stay reachable (TRA-794)

nikolai-vysotskyi:
Closes TRA-794.

## The problem

The first-run wizard's **Connect coding assistants** step renders a sheet as tall as its
content. `.lx-sheet-scrim` is `position: fixed` and nothing below it scrolls, so whatever
falls past the window edge is unreachable — and what falls off is the bottom, where the
actions are. A new user is shown a list of clients and no way forward; only Esc, which
means "skip".

Measured on the running renderer (`electron-cdp.mjs`, 10 MCP clients detected on this
machine, dev build of 3.16.0):

| | before | after |
|---|---|---|
| 960×700 — sheet height | 753 (window 700) | 676 |
| 960×700 — actions bottom edge | y 732 — off screen | y 656 |
| 640×420 (window minimum) — sheet height | 753 | 396 |
| 640×420 — actions bottom edge | y 732 — off screen | y 376 |
| list scrolls | no | yes |

## The fix

- `.lx-sheet` — `max-height: calc(100vh - 24px)` and a flex column, so a sheet can never
  exceed the window.
- `.lx-sheet-body` — `min-height: 0; overflow-y: auto`. The safety net for every step,
  including long translated copy at the window minimum.
- `.lx-sheet-actions` — `flex: none`, so the button row is never the thing that compresses.
- Clients step — the list card takes the leftover height (`flex-1 min-h-0`) and its rows
  scroll (`overflow-y-auto`). One scroll container, and it is the list: the title, subtitle
  and both buttons stay put.

## Verification

Screenshots before/after in light and dark and at the 640×420 minimum are attached to
TRA-794. `pnpm test` in `packages/app`: 65 files, 657 tests, all passing; `typecheck` clean.
The new test in `SetupWizard.test.tsx` fails if either half of the chain is dropped — the
CSS ceiling or the shrink/scroll classes on the list.

DESIGN.md §6 gains the general rule, since the missing ceiling was house-wide and not
specific to this step.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1182,6 +1182,25 @@ belongs to the window: the application menu, a context menu at the cursor.
 Sidebar collapsed, sidebar at 180 and at 320, and the 640px window minimum are four
 different pane boxes. Check the overlay in all of them, not just the one you developed in.
 
+### A sheet is clamped by the window, and its list scrolls — not the sheet
+
+`.lx-sheet-scrim` is `position: fixed`, so it cannot scroll: whatever part of a sheet falls
+past the window edge is gone, not merely below the fold. And the part that falls off is
+always the bottom, which is where the actions are. TRA-794 shipped that — with ten MCP
+clients detected, the first-run wizard's sheet measured 753px in a 700px window and both
+`Skip for now` and `Connect selected` sat at y 732. A new user got a list and no way
+forward; only Esc, which means "skip".
+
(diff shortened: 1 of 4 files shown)


nikolai-vysotskyi:
Duplicate of #856, which opened three minutes earlier with the same fix and more verification (Reduce Transparency / Increase Contrast, and the other three wizard steps re-measured). Closing this one; #856 is the PR for TRA-794.

PR #876, opened 2026-09-04
fix(docs): a narrow README banner so a phone stops rendering it at 8px (TRA-827)

nikolai-vysotskyi:
Closes TRA-827.

## The problem

GitHub renders the 1200 CSS px banner in a ~390 px column on a phone. That 0.33 scale puts the 25px tagline at 8px and the receipt figures below legibility — reported with a screenshot from the GitHub iOS app.

The buttons are fine: `width="250"` is already narrower than a phone's column, so they wrap into a stack at their full size. Only the banner breaks.

## The fix

`<source media>` is the only responsive lever GitHub's markdown sanitiser leaves us, and it takes any media query — not just `prefers-color-scheme`. The banner now also ships a 480 CSS px one-column cut behind `(max-width: 500px)`, listed **above** the theme-only source because the first match wins.

- `scripts/gen-readme-banner.mjs` — a `.narrow` modifier that restates only the sizes that break at 480px. Palette and copy are inherited, so a colour or wording change lands in both cuts.
- `README.md` — two `max-width` sources above the theme-only one, in the banner's `<picture>` only.
- `docs/images/readme/banner-narrow-{light,dark}.png` — new.
- `docs/DESIGN-WEB.md` §10 — why two cuts, and why the buttons don't get one.

## Test evidence

New `tests/docs/readme-header-images.test.ts` (3 tests): every referenced header asset exists on disk, the banner covers both themes at both widths with no stray button cut, and the `max-width` sources sit above the theme-only one. A broken `srcset` otherwise falls through silently — the desktop reviewer sees nothing wrong while the phone gets the wide cut.

```
npx vitest run tests/docs/      → 17 files, 222 passed, 4 skipped
pnpm run lint                   → clean
biome ci <changed files>        → clean
```

**Desktop is untouched, byte-for-byte.** After regenerating the whole set:

```
$ git diff --stat origin/master -- docs/images/readme/
(empty)
```

Rendered in headless Chrome at 390×844 @3x (dark) and 1440×900 @2x (dark) against a GitHub-like container; the wide layout is pixel-identical to master and the narrow one reads at native size.

Agent: Lead Engineer


nikolai-vysotskyi:
Closing in favour of #877 (reopened). Both PRs carried the same markup, and Code Reviewer found a defect in it that this branch shares: GitHub's `<themed-picture>` substring-matches `(prefers-color-scheme: light)`, so with Appearance pinned to Light it rewrites the compound narrow-light source to match every viewport — a 1440px desktop then renders the phone cut at 750px. The fix and the guarding test are going onto #877, which is where the review verdict is.

PR #877, opened 2026-09-04
fix(docs): a narrow banner cut so the README header stays readable on a phone (TRA-827)

nikolai-vysotskyi:
GitHub scales the README banner into the column it has. On a phone that column is ~324 CSS px, so the 1200px cut arrives at 0.27 scale: the 25px tagline lands at ~7px and the receipt is a smudge (see TRA-827, Nikolai's screenshot). The buttons are fine — at `width="250"` a plate is already narrower than the column, so the three just stack at full size.

`<source media>` is the only responsive lever GitHub's markup sanitiser leaves in a README, and it accepts any media query, not only `prefers-color-scheme`. So the banner now also ships at 480 CSS px in one column, picked by two `(max-width: 500px)` sources placed **above** the theme source — first match wins.

- `scripts/gen-readme-banner.mjs` — a `.banner.narrow` modifier overriding only the sizes that break at 480px; palette and copy stay shared, so a colour or wording change lands in both cuts.
- `README.md` — two `max-width` sources inside the existing `<picture>`.
- `docs/images/readme/banner-narrow-{light,dark}.png` — new, 77/76 KB after pngquant. Only one is ever fetched.
- `tests/docs/readme-header-images.test.ts` — all referenced header images exist, both cuts in both appearances, narrow sources above the theme source.
- `docs/DESIGN-WEB.md` §10 — why two cuts, and why the buttons don't need one.

## Verification

Rendered by GitHub itself on this branch (not a local Markdown preview), headless Chrome:

- 390×844 @3x — `img.currentSrc` = `banner-narrow-dark.png`, laid out at 324px. Readable.
- 1440×900 @2x — `img.currentSrc` = `banner-dark.png`, laid out at 750px. Unchanged.

The GitHub API's sanitised HTML for this branch keeps both `media` attributes verbatim, including the compound `(max-width: 500px) and (prefers-color-scheme: light)`.

Desktop assets are untouched: `git diff master -- docs/images/readme/` shows only the two new files. `pnpm run test --run tests/docs/` — 17 files, 222 passed. `pnpm run lint` clean.

Unrelated drift noticed while here: README prose and the banner `alt` still say "177 tools" while `counts.yml` and the rendered banner say 178. Left alone — separate change.

Agent: Lead Engineer


nikolai-vysotskyi:
Duplicate of #876 — a parallel run of the same agent opened that one seven minutes earlier with the same fix and the same source ordering. Keeping #876.

nikolai-vysotskyi:
Duplicate of #876, which is the earlier PR with the identical six-file diff — two runs of mine landed the same work. Continuing on #876; the review verdict on TRA-827 applies there and is being addressed. Closing this one so the Duplicate PR Check clears.

Agent: Lead Engineer

PR #896, opened 2026-09-04
fix(ci): only a claimed issue counts as a duplicate, not a cited one

nikolai-vysotskyi:
`scripts/check-duplicate-pr.mjs` treated any `TRA-NNN` string anywhere in a PR
title or body as a claim on that issue. A PR that names a sibling issue to
explain how the fixes relate therefore collided with the PR that actually
implements it, and `check` went red for a reason unrelated to the code.

Four open PRs were red on this on 2026-09-04, none of them duplicates:

- **#874** (claims TRA-811) flagged against **#871** (claims TRA-809), because
  its body reads "the daemon restarts constantly (TRA-809) and never reaches
  the 5-minute window there".
- **#875** (claims TRA-813) flagged against **#871** for the same sentence
  pattern, and against #874.
- **#882** flagged against **#854** on TRA-791 by the same mechanism.

That cross-reference is the writing the PR template asks for. The guard
punished it.

## Change

New `claimedKeys(pr)` separates the issue a PR *claims* from ones it merely
*references*: the title, plus an explicit `Closes/Fixes/Resolves TRA-NNN` in
the body. `findDuplicatePrs` compares claimed against claimed.

The guard keeps doing its job. Both TRA-476 incidents (#598/#597, #613/#611)
named the issue in the title *and* in a `Closes` line, so they are still
caught — there is a test asserting the title-only half of that. The
`Follow-up to #NNN` escape hatch and the release-PR exclusion are untouched.

The loosening is deliberate and narrow: a PR that names its issue *only* in
prose ("Part of TRA-500"), with no title key and no closing keyword, now
claims nothing and is skipped. Every real duplicate we have seen named it in
the title.

## Test evidence

`tests/ci/duplicate-pr.test.ts` — 15 passed (was 8; the 8 pre-existing cases
are unchanged and still green). New cases are built from the four real PRs
above plus the title-only regression guard.

Ran against the live PRs, all four previously-flagged now pass:

```
#871: PASS  #874: PASS  #875: PASS  #882: PASS  #854: PASS
```

Full suite: `896 passed | 5 skipped (901)` files, `9949 passed | 24 skipped`
tests. `pnpm run lint` (tsc --noEmit) clean.

Closes TRA-856.

Agent: Lead Engineer

### DIFF
--- a/scripts/check-duplicate-pr.mjs
+++ b/scripts/check-duplicate-pr.mjs
@@ -32,6 +32,24 @@ export function isReleasePr(pr) {
 }
 
 /**
+ * Issues a PR *claims* — the ones it says it implements — as opposed to ones it
+ * merely references.
+ *
+ * TRA-856: sibling fixes explain how they relate ("the daemon restarts
+ * constantly (TRA-809) and never reaches the window there"), which is the
+ * writing the PR template asks for. Reading every key in the body as a claim
+ * made #871/#874/#875 collide with each other and #882 with #854 — four PRs
+ * red for citing context. A claim is the title, or an explicit closing keyword.
+ *
+ * @param {{title?: string, body?: string}} pr
+ */
+export function claimedKeys(pr) {
+  const closes = [...`${pr.body ?? ''}`.matchAll(/\b(?:close[sd]?|fixe?[sd]?|resolve[sd]?)\s+(TRA-\d+)\b/gi)]
+    .map((m) => m[1].toUpperCase());
+  return [...new Set([...issueKeys(pr.title), ...closes])];
+}
+
+/**
  * Open PRs that claim the same TRA issue as `pr` and that `pr` does not cite.
  *
  * @param {{number: number, title: string, body?: string}} pr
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Closing as a duplicate of #897, which was opened three minutes later by a concurrent run on the same issue (TRA-856) with the same `claimedKeys` design. This is exactly the failure mode the guard exists to catch, and the guard caught it — that part worked.

#897 is kept for one concrete reason: its closing-keyword regex reads a multi-issue line and an optional colon, where the regex here takes only the first key. `Closes TRA-596, TRA-597, TRA-598, TRA-599, TRA-600` is a real body from this repo's history — under the single-key form, four of those five claims become invisible to the guard, so a genuine duplicate on TRA-597…600 would pass.

The one test here that #897 lacked — a rival claiming the issue in its title alone must still be flagged — has been ported over, with attribution to this PR. Nothing else from this branch is lost.

No reflection on the work; it was a good fix, and it landed second by three minutes.

PR #903, opened 2026-09-04
fix(mirror): exempt spill reads, make the rewrite byte-deterministic (TRA-860)

nikolai-vysotskyi:
Closes TRA-860.

JetBrains Research published a paired A/B over 80 Claude Code tasks in which `rtk`, a token-saving utility advertising 60–90% savings, **raised task cost by 7.6%** — through a cache-busting penalty: rewriting what the model sees changes the prompt prefix, a $0.30/M cache read becomes a $3.75/M cache write. Our Read/Bash mirrors are described by the same sentence, and 94% of our corpus by volume is cache reads, so the claim was checked directly instead of being answered with our own −14% from TRA-749 (taken on 2–4-turn tasks, where the prefix never grows large enough to be worth invalidating).

## Finding

**The mirrors do not break the prompt cache. They cut cache writes by 95%.**

`updatedToolOutput` replaces a tool result *before* it is appended, so the rewritten bytes are always the newest message — behind the last cache breakpoint, in a region that has never been cached. Nothing already in the prefix is touched. `rtk` is not a hook: it is a command the agent invokes instead of the shell, so it changes which calls happen and in what order. The two share a description, not a mechanism.

Both halves were measured anyway.

**Corpus** — 2,746 sessions, 113,501 API requests, via `scripts/mirror-cache-continuity.mjs`:

| | requests | prefix breaks (<90% carried) | prefix mass re-written |
|---|---|---|---|
| no mirror | 101,921 | 1,793 (1.76%) | 1.84% |
| after a mirror rewrite | 41 | 1 (2.44%) | 0.40% |

**Live A/B** — 8 large `Read` calls, 9 turns, hook on/off, empty MCP config:

| arm | cost | cache_read | cache_write | breaks | min hit ratio |
|---|---|---|---|---|---|
| bare (n=3) | $1.030 / $1.027 / $0.778 | 1.19 M | 191.7 K | 1 | 0.0 |
| mirror (n=2) | $0.165 / $0.192 | 0.49 M | 8.7 K / 10.4 K | 0 | 1.0000 |

The only cache break in the whole experiment happened in the *bare* arm. There is no session length at which mirrors stop paying: the saving is `(T−T′)·(W + R·Rd)` and grows with `R`. The break-even that does exist is the spill-recall rate, which has to exceed the compression ratio (90.4%) to hurt; corpus recall is 0 of 48.

Method, threats and the full break-even table: `benchmarks/mirror-prompt-cache.md`.

## Changes

The measurement found two real defects, fixed here:

1. **The escape hatch did not work.** The footer says `Full output: <spill>`, but a `Read` of that path was itself mirrored — verified live on a 41 KB spill, which came back as the same 1,669-char window plus a wasted turn and a second spill file. Reads under `TRACE_MCP_MIRROR_HOME` now pass through untouched. This was the only way this hook has to lose money.
2. **The rewrite was not byte-deterministic.** The spill was named `<epoch>-<pid>.txt`, so the same output produced a different replacement each run — the non-determinism TRA-858 is removing from our own tool outputs. Spill names are now content-addressed (sha256 prefix), which also stops the same output being spilled twice.

`MIRROR_HOOK_VERSION` 0.2.0 → 0.3.0. No change to the hook's output envelope, to any MCP tool signature, or to on-disk schema.

## Test evidence

Two new tests in `tests/hooks/mirror.test.ts` — one asserts a `Read` of a spill passes through, one asserts two identical outputs rewrite to identical bytes and produce one spill file, not two.

`pnpm test`: **904 files passed, 10,022 tests passed, 22 skipped**. `pnpm run build` clean, `biome check` clean.

Agent: Lead Engineer

### DIFF
--- a/benchmarks/e1-read-bash-mirror.md
+++ b/benchmarks/e1-read-bash-mirror.md
@@ -134,6 +134,13 @@ mirrors) over 10–15 live tasks with repeats. Nothing here says what compressio
 costs in correctness, re-asks, or call-count inflation — and the issue's own
 threat note is right that cheaper reads may simply buy more reads.
 
+## Does the rewrite break the prompt cache?
+
+No — it cuts cache writes by 95%. `updatedToolOutput` only ever alters the
+message being appended, never one already in the prefix, so no cached prefix can
+be invalidated by it. Measured on 113,501 requests of corpus and a live A/B:
+`benchmarks/mirror-prompt-cache.md` (TRA-860).
+
 ## Revised gate
 
 - ~~Death: adoption < 30%~~ — retired. Adoption is structural under the hook
--- a/benchmarks/mirror-prompt-cache.md
+++ b/benchmarks/mirror-prompt-cache.md
@@ -1,0 +1,141 @@
+# Read/Bash mirrors and the prompt cache (TRA-860)
+
+JetBrains Research published a paired A/B over 80 Claude Code tasks (July 2026)
+in which `rtk`, a token-saving utility advertising 60–90% savings, **raised task
+cost by 7.6%**. The mechanism reported was a cache-busting penalty: rewriting
+what the model sees changes the prompt prefix, a cache read at $0.30/M turns
+into a cache write at $3.75/M, and the 12.5× tariff eats the saving.
+
+Our mirrors (`hooks/trace-mcp-mirror.sh`) rewrite the output of native `Read`
+and `Bash` on the way to the model. That is the same sentence, so it deserved a
+direct measurement rather than a reference to our own −14% in TRA-749 — which
+was taken on 2–4-turn tasks, where the prefix never grows large enough to be
+worth invalidating.
+
+Measured 2026-09-05, Claude Code 2.1.239, `claude-sonnet-4-5`, macOS.
+
+## Answer
+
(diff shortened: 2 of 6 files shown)


nikolai-vysotskyi:
Both findings fixed in ddaa0138 — thanks, both were real.

1. **Hash fallback chain.** Each candidate now gets its own pipe, so an empty-pipe hash of the empty string cannot be produced at all. That is stronger than tightening the `-z` guard: the guard would still have to enumerate every hasher's empty-input output.
2. **Trailing slash.** `HOME_DIR="\${HOME_DIR%/}"` at assignment. Covered by a new test that fails without the strip (verified by reverting the one line).

On the `"$HOME_DIR"/*/*.txt` suggestion for the exemption pattern — leaving it broad. A user who points `TRACE_MCP_MIRROR_HOME` at `$HOME` has already told us to treat that tree as ours, and the tighter pattern would start compressing spills again the moment the layout under the home dir changes. Losing compression on a path inside our own state dir costs a few hundred tokens; compressing the escape hatch costs a wasted turn and a duplicate read.

`pnpm test` green (10,023 passed), biome clean.

Agent: Lead Engineer

PR #919, opened 2026-09-05
fix(test): unbreak master — formatting and a platform-dependent PID (TRA-881)

nikolai-vysotskyi:
master has been red since #916. Two failures, both in that PR's tests, neither in its shipped logic — `reconcileStaleBackups` is correct as written.

### 1. biome (reproduces on a clean master checkout)

`tests/init/launcher-integration.test.ts` was committed unformatted. `biome format --write`, no behaviour change.

### 2. `tests/updater-rollback.test.ts` — green on macOS, red on Linux

The case *"restores the newest stale backup when several dead-PID backups exist"* names its newer backup `bak-999`, and the whole case depends on that PID being **dead** so the backup counts as a restore candidate.

On Linux PID 999 is typically a root-owned system process. `process.kill(999, 0)` then raises `EPERM`, and `isPidAlive` counts `EPERM` as alive on purpose — EPERM means the process exists, we just cannot signal it (`src/updater.ts:158-168`). The backup was skipped as "owned by another updater", only the older one was left, and the assertion got `'stale-old-build'`. macOS leaves 999 free, so the case passes locally.

Fixed by using `999999998`, the same guaranteed-dead range the neighbouring case already uses. `'trace-mcp.tmcp-bak-4242'` still sorts first by name and last by mtime, so the case still proves mtime order beats readdir order.

### Evidence

```
npx vitest run tests/updater-rollback.test.ts tests/init/   18 files, 279 passed
npx biome ci .                                              1783 files, no fixes applied
```

Found while CI on #917 failed on changes that touch only markdown.

### DIFF
--- a/tests/init/launcher-integration.test.ts
+++ b/tests/init/launcher-integration.test.ts
@@ -290,16 +290,14 @@ describe.skipIf(process.platform === 'win32')('launcher shim integration', () =>
     // its mtime is almost always the newest thing in the root — and a cli.js
     // that exists there may still be mid-write. A complete backup wins even
     // when it is older (TRA-881).
-    it('prefers a complete backup over npm\'s newer in-progress staging dir', () => {
+    it("prefers a complete backup over npm's newer in-progress staging dir", () => {
       const { home, traceHome, node } = setupFakeHome();
       const root = path.join(home, '.nvm', 'versions', 'node', 'v22.22.2', 'lib', 'node_modules');
       for (const name of ['trace-mcp.tmcp-bak-4242', '.trace-mcp-deadbeef']) {
         fs.mkdirSync(path.join(root, name, 'dist'), { recursive: true });
         fs.writeFileSync(path.join(root, name, 'dist', 'cli.js'), `// ${name}\n`);
       }
-      const bakCli = fs.realpathSync(
-        path.join(root, 'trace-mcp.tmcp-bak-4242', 'dist', 'cli.js'),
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Closing: redundant, and my version is worse than what already landed.

When I branched, `origin/master` was at `1912101d` and both failures were real there — `biome ci` failed on `tests/init/launcher-integration.test.ts` from a clean checkout of that commit, and `tests/updater-rollback.test.ts` still used `bak-999`. Master has since moved to `f41969de` and both are fixed:

- `git show origin/master:tests/init/launcher-integration.test.ts` → `biome ci` exits 0.
- The backup case now uses `bak-999999997` / `bak-999999998`.

That second one is the part worth recording. Master avoids a low PID for **both** backups; my patch kept `bak-4242` as the older one and only fixed the newer. Same hazard, half addressed: if 4242 is live on a runner, `isPidAlive` reads EPERM as alive, `reconcileStaleBackups` skips it as owned by another updater, it never gets deleted, and `expect(listBackups()).toEqual([])` fails. Merging this would have traded one flake for another.

The diagnosis in the PR body still holds and is worth keeping findable: the original failure was PID 999 being a live root-owned process on the Linux runner, EPERM rather than ESRCH, and `isPidAlive` counting EPERM as alive is correct POSIX behaviour, not a bug. `reconcileStaleBackups` was never at fault.

No action needed — master is green.

PR #942, opened 2026-09-05
fix(app): four craft defects in the icon master (TRA-780)

nikolai-vysotskyi:
Follow-up to #925. Four craft defects Lead Engineer found in `assets/icon/icon.svg`, all measured, none touching the idea or the composition. Same mark, same nine nodes, same twelve edges. Release assets untouched.

## 1 — the `T` was a lighter patch, not a hole

It was filled solid `#6159E8`, sitting on top of the white hub. Rendered through librsvg and sampled:

```
T stem                    (512,560): rgb(97, 89, 232)
bare plate, same gradient  (150,892): rgb(78, 71, 229)
```

**19 levels lighter on R and G** — the letter glowed faintly instead of reading as a cut. Reproduced independently before touching it.

Now punched out of the hub with `fill-rule="evenodd"` on a compound path, so the letter's colour *is* the plate's gradient by construction and stays right if the gradient ever changes.

**Acceptance, after the fix**, both samples at the same gradient position (`x+y = 1072`):

```
T stem     (512,560): rgb(77, 71, 229)
bare plate (180,892): rgb(77, 71, 229)     delta 0, 0, 0
```

Holds in the SVG and in the PNGs the build script emits at 1024 and 512.

A consequence worth noting: edges used to run to the hub's *centre*, so a hole would have exposed white line ends. They now stop at the hub's rim — less overdraw, and nothing white behind the letter.

## 2 — node radii encoded nothing

Seven arbitrary values inherited from the raster. The degree-1 leaves held both the largest radius (44) and the smallest (26); five degree-3 nodes carried five different sizes (36, 36, 32, 32, 30).

Now three steps keyed to degree in the graph:

| role | degree | r |
|---|---|---|
| hub | 6 | 124 |
| node | 3 | 36 |
| leaf | 1 | 24 |

Both satellite steps are multiples of 12, so they land on the pixel grid the same way at every power-of-two size.

## 3 — top gloss 15 % → 8 %

Nothing else in the dock carries one. Kept at a trace rather than removed; the three options are on the board.

## 4 — `T` cap 36 % → 44 % of the hub

The hub was reading as a large empty circle. Bar/stem ratio (0.215 / 0.225) left alone as instructed.

## Checks

- `vitest run` in `packages/app` — 66 files, **675 passed**, 1 skipped
- `tsc --noEmit` and `tsc -p tsconfig.main.json --noEmit` — both clean
- `node scripts/gen-app-icon.mjs` regenerates the full set; `.icns` and `.ico` parse back correctly
- Test file grew three guards: the `T` is a hole and `#6159E8` cannot come back, satellites carry at most two distinct radii, and the hub is a compound path

Nothing in `packages/app/build/` changed.


nikolai-vysotskyi:
Superseded by #949, closing to clear the duplicate-work guard.

Both PRs fix the same four craft defects on the same masters, found by the same review of #925. #949 is a strict superset: it carries the four fixes, plus the switch-over that finishes TRA-780 — `release.yml` now calls `gen-app-icon.mjs`, `build/` is regenerated from the vectors, and `generate-icons.mjs` (296 lines holding the artwork as coordinate literals) is deleted with its package script.

One substantive difference, not just scope: this PR punches the `T` out of the hub disc. The hub-to-bottom edge runs under the letter and shows through a hole in the disc alone, so the mask has to cover the whole mark. #949 does that.

Nothing here is lost — the assertions this PR added to `app-icon.test.ts` are in #949 as well. Sorry for the collision; the guard caught it before either landed, which is what it is for.

PR #956, opened 2026-09-05
perf(bench): add session-scaling benchmark harness and v3.18.0 baseline (TRA-931)

nikolai-vysotskyi:
## What

The session-scaling benchmark harness and the v3.18.0 baseline it produced. Opened by the Lead Engineer: the work was complete on `agent/tracemcp-research-analyst/1e92a2839b8f` but no PR was ever raised, so the baseline every other performance issue measures against was stranded on a branch and could not be re-run by anyone else.

## Why

TRA-921's findings were hand-taken `ps` snapshots and one-off `sample(1)` runs on a single machine. Enough to prove the bugs; not enough to prove a fix worked, and useless against regressions. The reported symptom only appears at 8–10 concurrent sessions, so a harness that measures one session would call the problem fixed while it is not.

## What it measures

`scripts/bench-session-scaling.ts` drives a real daemon plus a pool of stdio sessions in an isolated temporary git repo pinned to v3.18.0 (`9256cf18`, 898 TypeScript files, 11 134 symbols), at N = 1 / 4 / 9, in two arms — daemon healthy vs `TRACE_MCP_NO_DAEMON=1`:

- cold start to first tool response
- session RSS and thread count at 60 s idle
- CPU cost of a one-file change, split between daemon and sessions
- deltas against the previous run, so a future PR can show it did not regress

`docs/perf/prereg-session-scaling.md` registers the questions, metrics, corpus and pass bars before the numbers, per the TRA-920 discipline. `docs/perf/session-scaling.json` is the machine-readable baseline.

## What the baseline says

| N=9 | daemon healthy | daemon absent |
|---|---|---|
| total system RSS | 2 449 MB | **1 645 MB** |
| cold start | 3 608 ms | **2 164 ms** |
| RSS growth N=1 → N=9 | **3.64×** | 9.06× |

The daemon wins on growth rate and on keeping session reindex CPU at 0.00 s, and loses on absolute memory and first response. That result is what TRA-941 now tracks, and it is only visible because this harness exists.

## Caveat, stated in the docs

898 files is a small corpus. Field observations run several times higher. This baseline is a floor, not a description of a large repository — the docs say so, and the Performance Agent's instructions now forbid quoting a harness number as if it described a large project.

Closes TRA-931.

### DIFF
--- a/docs/perf/prereg-session-scaling.md
+++ b/docs/perf/prereg-session-scaling.md
@@ -1,0 +1,78 @@
+---
+layout: default
+title: Preregistration — concurrency & session scaling benchmark
+permalink: /perf/prereg-session-scaling/
+description: What the concurrency and session scaling benchmark set out to measure, the pass bars, and the control condition comparing daemon-healthy vs daemon-absent architectures.
+noindex: true
+measurement: session_scaling
+data_file: docs/perf/session-scaling.json
+preregistration: yes
+written_on: 2026-09-05
+verdict: MET
+---
+
+# Preregistration — concurrency & session scaling benchmark (TRA-931)
+
+**This measurement was preregistered.** Written on 2026-09-05 for TRA-931 (parent tracking issue TRA-921).
+Following the discipline established in TRA-920, the test plan, metrics, frozen corpus, pass bars, predictions,
+and control condition were declared and frozen before recording the v3.18.0 baseline dataset.
+
+## Question
+
+As concurrent stdio AI coding agent sessions scale (N = 1, 4, 9) working on a shared repository:
+1. What is the cold-start latency, peak RSS, and thread cost to first tool response (`get_project_map`)?
+2. What is the steady-state idle cost at 60s (RSS and thread count) comparing a healthy daemon versus an unmanaged local fallback?
+3. What is the CPU seconds and wall time consumed across the daemon and all sessions when an edit changes a single file?
+4. What is the exact scaling multiplier of system resource consumption at N=9 vs N=1 across both architectures?
+
+## Metric
+
+All metrics are measured from real process trees using system primitives (`ps -o rss=`, `ps -M`, `ps -o cputime=`):
+- **Cold start wall time:** Milliseconds elapsed from session spawn until JSON-RPC stdio response to `tools/call` (`get_project_map`) arrives.
+- **Cold start peak RSS & threads:** Peak resident set size (MB) and thread count observed during session initialization.
+- **Steady-state idle (60s):** Per-session RSS (MB), daemon RSS (MB), total system RSS (MB), and total thread count measured after a 60-second idle hold to permit garbage collection and event loops to reach steady-state.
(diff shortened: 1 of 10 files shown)


nikolai-vysotskyi:
## Verdict: APPROVE

Reviewed as Reviewer C (third reviewer, standing in while Reviewer B and Code Reviewer are both quota-blocked).

### What I checked

- `gh pr view/diff 956` — note: the PR's true diff is 4 commits / 10 files (confirmed via `gh pr view --json commits,files`). A local `git diff master...HEAD` initially looked much larger because the checked-out `master` ref was stale; re-diffing against `origin/master` (fetched fresh) matched GitHub's file list exactly.
- Checked out `agent/tracemcp-research-analyst/1e92a2839b8f`, ran the full affected suites: `tests/daemon/*` and `tests/cli/*` — **415 passed, 1 skipped, 0 failed**. `tsc --noEmit` — clean.
- Read the three production-code changes (this is nominally a "benchmark harness" PR but it also ships real fixes found while building the harness, disclosed in the PR body):
  - `src/global.ts` — `DEFAULT_DAEMON_PORT` now reads `TRACE_MCP_DAEMON_PORT` if numeric, else falls back to 3741. Lets the benchmark run N daemon instances without port collision. Checked all call sites (`cli.ts`, `cli/daemon.ts`, `daemon/lifecycle.ts`, `daemon/client.ts`) — all just consume the constant, no assumption broken.
  - `src/cli.ts` — `serve-http`'s `--port` default now derives from the same constant instead of a separate hardcoded `'3741'` literal. Pure de-duplication.
  - `src/daemon/router/session-db.ts` — `seedSessionDbFromShared` now refuses to seed from a shared DB with 0 rows in `files` (new `countIndexedFiles` helper, try/catch-safe against a missing table). Fixes a real bug: a freshly-registered project's shared DB exists before `indexAll` ever runs, so a daemonless session used to seed itself from that empty DB, get treated as "already indexed", and serve an empty index forever. New test covers exactly this case, and the existing "copies content" / "missing DB" tests still pass.
- Scanned `scripts/bench-session-scaling.ts` (1078 new lines) for the usual trust-boundary suspects: the one `execSync` with string interpolation (`git archive ${PINNED_COMMIT} | tar -x -C "${fixtureDir}"`) only ever receives a hardcoded commit SHA and a `mkdtempSync`-generated path — no attacker/user-controlled input reaches it. Everything else is `execFileSync` with argument arrays. Dev-only tooling, not reachable from any MCP tool surface.
- Confirmed this PR touches no MCP tool contracts and no on-disk index schema — token-budget and schema-migration concerns from the review checklist don't apply here.

### Findings

None. Minor, non-blocking observation (not filing as a finding): `Number(TRACE_MCP_DAEMON_PORT)` accepts whitespace-only strings as `0` and fractional strings as-is (e.g. `"3741.5"`), since `Number.isNaN` doesn't catch either. Internal/dev-facing env var, not worth gating this PR on.

### What I ran

- `npx vitest run tests/daemon/session-db.test.ts tests/cli/env-overrides.test.ts` → 18/18 passed
- `npx vitest run tests/daemon tests/cli` → 415 passed, 1 skipped
- `git diff origin/master...HEAD --stat` (post-fetch) to confirm true PR scope matches `gh pr view --json files`

Merging.

PR #972, opened 2026-09-05
fix(app): stop comparing KPI tiles against a baseline minutes old (TRA-958)

nikolai-vysotskyi:
Fixes the "No change vs 2 seconds ago" captions on the Workspace headline tiles (TRA-958).

## The defect

`WorkspaceHeader` rolls a KPI baseline on the first frame that has a real reading, then compares the current KPIs against it immediately. On a first launch — or in any sandbox with a wiped profile — the "previous" snapshot is seconds old, every delta is zero, and all three tiles print `No change vs 2 seconds ago`. That caption describes when the app was opened, not the workspace.

`scripts/capture-screenshots.mjs` captures light before dark in one profile, so the light pass writes the baseline and the dark pass photographs the artefact. `docs/images/app-dark-projects.webp` ships it today, next to a light frame of the same surface whose tiles read "tracking from today", "1,264 per project", "5 per file".

## The fix

Fix direction (1) from the issue, in `kpiBaseline.ts`: a snapshot younger than `BASELINE_MIN_AGE_MS` (1 hour) is not history yet. `rollBaseline` returns no delta for it and keeps it stored rather than re-stamping `at` — an overwritten timestamp would never mature. Tiles with no delta already fall back to their footnote (`KpiTile`), which is a statement about the workspace.

This fixes the real app on the day a user installs it, not just the capture harness.

## Tests

`packages/app/src/renderer/workspace/__tests__/kpiBaseline.test.ts` — new case: no delta against a 2-second-old snapshot, and it is not re-stamped. Two existing cases that used a 1-minute-old baseline to mean "stored today" now use `BASELINE_MIN_AGE_MS + 1`.

App suite: 70 files, 702 tests passed. `tsc --noEmit` clean. The root suite excludes `packages/app` and is untouched by this diff.

## Not in this PR

Re-running the capture script and re-committing `docs/images/`. The script activates the app and owns the screen for the run (it says so in its own header, TRA-403), and the workspace rule is that a run must not seize the screen while someone is at the keyboard. Left to a Design/UX capture run after this lands.

Agent: Lead Engineer

### DIFF
--- a/packages/app/src/renderer/workspace/__tests__/kpiBaseline.test.ts
+++ b/packages/app/src/renderer/workspace/__tests__/kpiBaseline.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { BASELINE_MAX_AGE_MS, rollBaseline } from '../kpiBaseline';
+import { BASELINE_MAX_AGE_MS, BASELINE_MIN_AGE_MS, rollBaseline } from '../kpiBaseline';
 import { relativeTime } from '../../i18n/format';
 import type { WorkspaceKpis } from '../types';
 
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
Superseded by #974 (TRA-960), which landed the same one-hour floor on `BASELINE_MIN_AGE_MS` while this was in review. Master's version is equivalent — the age check sits after the 24h roll rather than before it, and since 1h < 24h the branches are the same. Closing rather than rebasing onto an identical fix.

PR #977, opened 2026-09-05
fix(app): stop the KPI strip comparing the workspace to itself (TRA-960)

nikolai-vysotskyi:
Fixes TRA-960.

Measured on the running Electron window (`scripts/electron-cdp.mjs`), a first launch read:

```
Projects | 18    | ↑ +3      vs 34 seconds ago
Files    | 23.2k | No change vs 34 seconds ago
Symbols  | 166.1k| No change vs 34 seconds ago
```

Three defects in one line of the card:

1. **The comparison compares the workspace to itself.** `rollBaseline` showed a stored snapshot the moment it existed — the only floor was the 24h `BASELINE_MAX_AGE_MS` ceiling. Every new install saw "No change vs 34 seconds ago" in the slot DESIGN.md §7 reserves for the answer to "compared to what?".
2. **The app's own load sequence was reported as workspace growth.** `↑ +3` was three projects that finished loading *after* the baseline was snapshotted. TRA-458's defect one step in: that one refused an *empty* snapshot, this refuses a *just-taken* one.
3. **The caption froze at mount.** `deltaCaption` was `useMemo`'d on `[baseline, t]` and `baseline` is set once per mount, so `Date.now()` was read once. The delta *number* kept tracking the 2s poll while the caption beside it did not — a window left open all day reported a whole day of growth as the last half-minute's.

## Fix

- `BASELINE_MIN_AGE_MS = 1h` in `kpiBaseline.ts`: a snapshot younger than an hour is stored but not compared against. Kills 1 and 2 with one rule; the tile falls back to the no-baseline state it already has.
- `deltaCaption` computed at render instead of memoized. The pane re-renders on its own 2s poll, so no timer is added.

## Verified on the running app

Fresh profile, no baseline — the designed fallback, not a tautology:

```
Projects 19 / tracking from today
Files    24.7k / 1,298 per project
Symbols  175.6k / 7 per file
```

Real comparison path intact, with a baseline planted 5 hours back:

```
Projects 22    ↑ +9    vs 5 hours ago
Files    24.7k ↑ +1.5k vs 5 hours ago
Symbols  175.6k ↑ +30.9k vs 5 hours ago
```

Screenshotted in both appearances; card geometry unchanged — `KpiTile` reserves the comparison line's 26px whatever the string does.

`DESIGN.md` gains the rule next to the empty-snapshot one it is a twin of.

Locally: 705 tests pass, `tsc --noEmit` clean, `check-i18n` clean, `pnpm build` green.


nikolai-vysotskyi:
Duplicate of #974, which landed the byte-identical change while this run was in flight (`git diff origin/master HEAD` on this branch is empty). Closing in favour of it.

PR #981, opened 2026-09-05
fix(app): the Index card's stale line stops asserting a busy daemon (TRA-964)

nikolai-vysotskyi:
Fixes TRA-964.

## What was on screen

Project Overview's Index card, on the running Electron window (app 3.18.0), a real project:

```
Index
  The daemon is busy. These are the last indexed numbers.
  Status                                   ● Daemon unreachable
  Files indexed                                       6,641
```

A poll later, same card: the same first line above `Status  Not tracked`. Two verdicts about one daemon, 48px apart.

## Why

`ProjectOverview.tsx` rendered `t('workspace:busyStale')` whenever `!statsFresh`. Only the second half of that string — "These are the last indexed numbers." — is what `!statsFresh` knows. The first half is a diagnosis, and on the Workspace it is *chosen*: `busyMessage()` picks between four sentences from a real busy reading. Project Overview asserted it for every stale reading, directly above the row whose whole job is to name the cause.

## Change

New `overview:staleNumbers` in all ten catalogues — each locale's string is the second half of its own existing `busyStale`, so nothing needed re-translating. The card now reads one sentence about the numbers and one row about the daemon, with no combination in which they can disagree.

Verified on the running window in both appearances (`scripts/electron-cdp.mjs`, stats endpoint blocked so the card falls back to its stored snapshot):

```
Index
  These are the last indexed numbers.
  Status                                   ● Daemon unreachable
```

The regression test asserts the card cannot print "daemon is busy" while its Status row reads "Not tracked"; confirmed to fail against the old code.

`DESIGN.md` gains the rule this is an instance of: a string borrowed from another surface carries that surface's verdict.

Local: `pnpm test` 710/710, `typecheck`, `check:i18n` clean. Reviewed by a second model.

Agent: Design/UX Agent

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -675,6 +675,16 @@ statement: what could not be loaded, plus the control that tries again.** Retry 
 the next step and it is already in the row; the invented sentence between them only
 had to be wrong once to cost the reader a diagnosis.
 
+**A string borrowed from another surface carries that surface's verdict.** Project
+Overview's Index card printed the Workspace's `busyStale` — "The daemon is busy. These
+are the last indexed numbers." — whenever its numbers came from the stored snapshot.
+Only the second half is what "stale" knows; the first half is a diagnosis, and on the
+Workspace it is *chosen* from a real busy reading. Here it was asserted 48px above the
+Status row whose whole job is to name the cause, which read "Daemon unreachable", then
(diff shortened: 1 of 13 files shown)


nikolai-vysotskyi:
Duplicate of #980, which is the same fix (same `overview:staleNumbers` key, same ten locales, same regression test) and already CI-green. Closing this one; #980 is the merge.

PR #985, opened 2026-09-06
fix(config): cap unregistered project sections in .config.json (TRA-706)

nikolai-vysotskyi:
Closes TRA-706 (split out of TRA-701; TRA-702 landed the sweep itself).

## Problem

TRA-702 gave `.config.json` its first collector — a section is dropped when its
root is gone and no registry entry claims it, or when the root is a one-shot
Multica workdir. That removes 440 of the 591 sections on the reporting machine,
but both rules are heuristics *over* the file rather than a ceiling *under* it:
a runtime that scatters live scratch checkouts under any other path shape keeps
the map growing, and the whole file is reparsed on every server start.

## Change

After the two existing rules, the sweep keeps at most `MAX_UNREGISTERED_SECTIONS`
(100) sections that no registry entry claims, evicting oldest first. `modify()`
appends new keys and edits existing ones in place, so file order is insertion
order and the overflow slice cuts the right end. Registered projects are never
evicted — `registry.json` is already bounded by deliberate `add`/`init`.

At ~1.7 KB per section that is the stated ceiling the issue asked for: ~170 KB
for the unregistered half, against the 1 MB that prompted it. Documented in
`docs/configuration.md`.

## Tests

Two cases added to `src/__tests__/config-project-sweep.test.ts`:

- 105 live, non-workdir-shaped roots → the five oldest are evicted, the rest kept.
- a registered project written first, plus 101 scratch roots → only the oldest
  scratch root goes; the registered one survives past the cap.

`pnpm test`: `Test Files 926 passed | 3 skipped · Tests 10253 passed | 24 skipped`.
`pnpm run build` and `tsc --noEmit` clean.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,15 +58,21 @@ All trace state lives in `~/.trace/` (with automatic backwards compatibility and
     my-app-a1b2c3d4e5f6.db  # per-project databases
 ```
 
-### Pruning of per-project sections
+### Pruning of projects
 
-The server writes one `projects[...]` section per directory it indexes, and reparses
-every section on start. An hourly sweep drops the sections nothing needs: a root that
-no longer exists and no registry entry claims, and a one-shot agent-run workdir. On top
-of that, at most **100 sections not claimed by `registry.json`** are kept, oldest first
-— so the unregistered half of the file stays around 170 KB no matter how many throwaway
-directories have been indexed. Projects added with `trace add` / `trace init` are
-registered and are never pruned.
+Starting the server in a directory auto-registers it, so both `registry.json` and the
+`projects[...]` map in `.config.json` grow on their own — and both are read on every
+server start. An hourly sweep bounds them:
+
+- a root that no longer exists (after a 7-day grace, so an unmounted drive keeps its
+  registration) and one-shot agent-run workdirs are dropped;
+- at most **100 auto-registered projects** are kept, least recently indexed first;
+- at most **100 config sections that no registry entry claims** are kept, oldest first.
+
+Projects you registered yourself with `trace add` or `trace init` are exempt from the
+caps and are only removed by `trace remove` / `trace-mcp doctor --fix`. Eviction only
+deregisters — your directory and its index database are left alone, so an evicted
+project is re-registered the next time you open it.
 
 ### Config merge order
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,6 +58,16 @@ All trace state lives in `~/.trace/` (with automatic backwards compatibility and
     my-app-a1b2c3d4e5f6.db  # per-project databases
 ```
 
+### Pruning of per-project sections
+
+The server writes one `projects[...]` section per directory it indexes, and reparses
+every section on start. An hourly sweep drops the sections nothing needs: a root that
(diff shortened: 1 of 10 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Caps unbounded growth of `registry.json` and `.config.json`'s `projects` map (TRA-706), by distinguishing deliberate registration (`trace add`/`trace init`, now tagged `explicit: true`) from implicit auto-registration (`serve` on whatever directory it's started in), and evicting only the implicit class, oldest-first, past a 100-entry cap on each store.

### What I checked
- Traced `registerProject`'s branching by hand for the cases that matter:
  - Existing + no `type`/`children` in `opts` → promotion branch (`explicit && !existing.explicit && !ephemeral` → mutate in place + `saveRegistry`) fires correctly whether `opts` is `undefined` (plain re-registration) or `{explicit: true}` (a later `trace add`) — confirmed by the "later `trace add` promotes an already auto-registered project" test.
  - `isGeneratedSection`'s comment detection uses the parser's own comment-range list rather than scanning source text — the file's own comment explains a prior draft matched `**/` inside `"include": ["src/**"]` as a comment; the current approach avoids that class of bug entirely, and the "never evicts a section a human has touched" test exercises exactly that fixed case.
  - `pruneProjectConfigSections`'s overflow cap (`keptUnregistered.slice(0, overflow)`) evicts oldest-first by insertion order, matching `modify()`'s append-at-end behavior — confirmed by the dedicated ordering test.
- One low-severity edge case I traced but am not filing as a blocker: if the *same process* first registers a workdir-shaped root implicitly (landing in the process-local `_ephemeralEntries` map) and then explicitly re-registers it (`opts.explicit: true`), the lookup switches to `reg.projects[absRoot]` (since `ephemeral` becomes `false` once `opts` is truthy) and won't find the entry there, so it re-registers from scratch with a different `dbPath` (`getDbPath` vs `getEphemeralDbPath`) instead of promoting the existing one. In practice this can't happen: `_ephemeralEntries` is process-local and `trace add`/`trace init` are separate CLI invocations from the long-running `serve` process that would have created the ephemeral entry, so there's no process where both paths run against the same map. Worth a code comment if anyone extracts this into a shared helper later, not worth blocking on now.
- Ran the full set of tests that exercise `registerProject`/`setupProject`/`pruneProjectConfigSections` and their downstream consumers (registry health, doctor, remove, ask-resolve, overlap/ephemeral/cleanup project-setup suites, daemon project-manager, watcher lifecycle) — **135 + 43 = 178 passed, 0 failed**.
- `npx tsc --noEmit` — clean.

### Findings
None blocking (see the traced-but-unreachable edge case above, noted for awareness only).

### What I ran
- `npx vitest run src/__tests__/config-project-sweep.test.ts tests/project-setup/implicit-project-cap.test.ts` → 12 passed
- `npx vitest run src/__tests__/{db-holders,git-remote-identity,guard-init,pending-reindex-attempts,registry-health,config-jsonc-merge}.test.ts tests/cli/{ask-resolve-remote-repo,doctor-registry,remove}.test.ts tests/project-setup/ tests/registry.test.ts` → 135 passed
- `npx vitest run tests/daemon/project-manager-*.test.ts tests/indexer/watcher-lifecycle.test.ts` → 43 passed

Merging.

PR #986, opened 2026-09-06
fix(server): stop an orphaned server spinning forever on a dead log pipe (TRA-921)

nikolai-vysotskyi:
## What

An orphaned `serve-http` / `serve` process spins at ~100% CPU forever and never releases its port.

## Measured

A `serve-http` daemon left over from an earlier benchmark run (another agent's workdir, port 37497) was found holding **517 MB and 97% of a core with zero clients, 5h56m after its parent exited** — 356 minutes of CPU spent on nothing.

`sample(1)` on it, and again on a fresh reproduction, showed the same loop and **no SQLite frames at all**:

```
uv__run_check
  node::Environment::CheckImmediate
    v8::internal::Isolate::ReportPendingMessages
      node::errors::TriggerUncaughtException
        v8::internal::Accessors::ErrorStackGetter
          v8::internal::ErrorUtils::FormatStackTrace   <- and round again
```

## Mechanism

When the parent that owned our stdout/stderr pipe exits without killing us, every write to those fds fails with `EPIPE`. `installProcessSafetyNet`'s `uncaughtException` handler exists to keep a long-lived server alive through stray errors, and it does that by *logging* — to the same dead pipe. That write throws `EPIPE`, which re-enters the handler, which formats a stack trace (`formatErr` reads `err.stack`), which logs, forever.

## Reproduction

Deterministic, ~6 seconds:

```js
const child = spawn('node', [CLI, 'serve-http', '-p', PORT], {
  stdio: ['ignore', 'pipe', 'pipe'], detached: true,
});
child.unref();
setTimeout(() => { child.stdout.destroy(); child.stderr.destroy(); }, 6000);
```

## Before / after

Same machine (M5 Max), same repro, 17 s after the pipes close:

| Orphaned `serve-http` | Before | After |
|---|---|---|
| CPU | 88–106% | process has exited |
| RSS | 853 MB | 0 |
| Lifetime | unbounded | ends at the first `EPIPE` |

## The change

A dead log sink is the one error the safety net must not swallow — there is nowhere left to report to, so the process leaves. Ordinary uncaught exceptions still log and keep the process alive, unchanged.

- `isBrokenLogSink()` recognises `EPIPE` / `ERR_STREAM_DESTROYED` / `ERR_STREAM_WRITE_AFTER_END`.
- Regression guard in `src/server/__tests__/process-safety-net.test.ts`: an EPIPE `uncaughtException` exits, an ordinary one does not. Fails without the fix.
- Finding written up in `docs/perf/idle-cost.md`.

No MCP tool contract change, no token cost change, no schema change.

Local: full suite 10 255 passed / 24 skipped, biome clean, build green.

Closes part of TRA-921.

### DIFF
--- a/docs/perf/idle-cost.md
+++ b/docs/perf/idle-cost.md
@@ -65,3 +65,47 @@ The before/after gap is dominated by lock-queue wait, which is also why `reindex
 telemetry used to report elapsed times in the hours: `elapsedMs` summed the wait with the work.
 It is now the work alone, with the wait beside it as `queuedMs` (`daemon stats` renders it as
 `queued p95`).
+
+## The most expensive idle process is the one nobody owns
+
+Measured 2026-09-06 on the M5 Max. A `serve-http` daemon left over from an earlier benchmark run
+(port 37497, in another agent's workdir) was holding **517 MB and 97% of a core, with zero
+clients, 5h56m after its parent exited** — 356 minutes of CPU spent on nothing.
+
+`sample(1)` on it, and again on a fresh reproduction, showed the same loop and no SQLite at all:
+
+```
+uv__run_check
+  node::Environment::CheckImmediate
+    v8::internal::Isolate::ReportPendingMessages
+      node::errors::TriggerUncaughtException
+        v8::internal::Accessors::ErrorStackGetter
+          v8::internal::ErrorUtils::FormatStackTrace   <- and round again
+```
+
+The mechanism: when the parent that owned our stdout/stderr pipe exits without killing us, every
+write to those fds fails with `EPIPE`. The `uncaughtException` safety net in
+`src/server/process-safety-net.ts` exists to keep a long-lived server alive through stray errors,
+and it does that by *logging* — to the same dead pipe. That write throws `EPIPE`, which re-enters
+the handler, which formats a stack trace, which logs, forever. The process never exits and never
+releases its port.
+
+Reproduction, deterministic in about six seconds:
+
+```js
+const child = spawn('node', [CLI, 'serve-http', '-p', PORT], {
+  stdio: ['ignore', 'pipe', 'pipe'], detached: true,
+});
+child.unref();
+setTimeout(() => { child.stdout.destroy(); child.stderr.destroy(); }, 6000);
+```
+
+| Orphaned `serve-http`, 17 s after the pipes close | Before | After |
+|---|---|---|
+| CPU | 88–106% | process has exited |
+| RSS | 853 MB | 0 |
+| Lifetime | unbounded | ends at the first `EPIPE` |
+
+The fix is to treat a dead log sink as the one error the safety net must not swallow: there is
+nowhere left to report to, so the process leaves. Guarded by
+`src/server/__tests__/process-safety-net.test.ts`.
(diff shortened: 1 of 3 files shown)


nikolai-vysotskyi:
Merge gate: both required reviewers are quota-exhausted right now — Reviewer B returned `provider_quota_limit` with a ~115h reset, Code Reviewer has failed four tasks in a row on `You've hit your usage limit`. The workspace's degradation rule only allows an author-merge after both have been unavailable for four continuous hours; that threshold is not met yet, so this stays open and green rather than being merged by its author. One APPROVE from either reviewer closes it.

nikolai-vysotskyi:
Revised in `b1a21eae` after review. The first cut exited on a generic error code reaching the global `uncaughtException` handler, which would have taken a healthy daemon down over an unrelated socket's EPIPE. Instrumenting the reproduction showed the error is emitted as an `error` event on `process.stderr` and only reaches the global handler because nothing listens — so the guard now lives on `process.stdout`/`process.stderr` themselves, and `uncaughtException` is back to log-and-survive with no exit branch. Tests assert both directions; the negative case fails against the previous implementation. CI green.

Gate status unchanged: Reviewer B is in quota until ~Sep 10, Code Reviewer is failing intermittently on usage limits (it did land the review above, then went back to failing). Not merging by author — the workspace rule needs four continuous hours with both unavailable and that threshold is not met. One APPROVE from either closes this.

nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Fixes an orphaned `serve-http` spinning forever burning CPU when its stdout/stderr pipe dies (parent exits without killing it): every write fails EPIPE → uncaught exception → safety net logs → logs to the same dead pipe → repeat forever. Fix listens on `process.stdout`/`process.stderr` `error` directly and exits on a broken-pipe code from those specific streams.

### What I checked
- Read `src/server/process-safety-net.ts` in full. Confirmed `guardLogSinks()` is only ever registered once per process — it's called at the end of `installProcessSafetyNet`, which is itself guarded by the top-level `installed` flag, so no duplicate-listener risk from repeated calls.
- Checked both entry points call it: `src/cli.ts:412` (`serve`), `src/cli.ts:591` (`serve-http`) — both covered.
- Verified there's no pre-existing `process.stdout`/`stderr` `'error'` listener elsewhere in `src/` that this would conflict with or double-fire against.
- Confirmed the fix is scoped correctly: `isBrokenPipe()` is only consulted for errors the stream itself raised (via the `stream.on('error', ...)` closure), so the same EPIPE code from an unrelated HTTP response or socket — routed through `uncaughtException` instead — never triggers the exit path. The new test `does NOT exit on the same code from an unrelated stream` pins exactly this.
- In stdio mode, `process.stdout` is the actual JSON-RPC transport, not just a log sink (contradicting the local comment "guard the two streams the logger writes to", which is stdio-mode-inaccurate) — but exiting when that pipe breaks is still correct: if the client's stdin/stdout connection is gone, there's nothing left to serve either.

### Findings
None (the "guard the two streams the logger writes to" doc comment undersells what it does in stdio mode, but the behavior itself is correct either way — not worth blocking on).

### What I ran
- `npx vitest run src/server/__tests__/process-safety-net.test.ts` → 11 passed

Merging.

PR #988, opened 2026-09-06
feat(doctor): validate the launcher path clients actually spawn

nikolai-vysotskyi:
`launcher.log` is written *by* the shim. Every failure that happens before the shim executes is invisible to it — missing path, dangling symlink, lost `+x`, a directory, a gone home. In all of those the client shows `Failed to connect` and the log stays clean at zero, which is exactly how TRA-910 went unnoticed for 24 hours while `doctor` reported all green.

This adds a check of the file a client actually spawns.

- `checkLauncherFile(path)` — lstat, resolve a symlink, require a regular file that is executable and carries our launcher header. One specific status per failure: `missing`, `dangling_symlink`, `not_a_file`, `not_executable`, plus `foreign` (a user's own wrapper — reported, not a breakage) and `unchecked` (a bare command name the client resolves through PATH).
- `checkRegisteredLaunchers()` — runs that check against the `command` every MCP client config registers for us, reusing the conflict detector's config-path list. Reads both the `trace` and legacy `trace-mcp` keys, and the per-project copies inside `~/.claude.json` that a running client actually spawns. Always includes the installed launcher path so a machine with no client configured still gets told its shim is broken.
- `trace doctor --launcher` prints a `Registered paths` section and exits non-zero on a broken one, with an outro that names the empty-log trap.
- `installLauncher()` no longer treats "header version matches" as healthy. A shim that lost its execute bit reads current and was skipped by every later `trace init`; now it is reinstalled. The dangling-legacy-symlink repair (#937) already covers the other half.

## Tests

`tests/init/launcher-health.test.ts` — 16 tests: each failure mode of `checkLauncherFile`, the registered-path scan (legacy key, per-project entries, malformed/absent configs, dedup across clients), and the install self-heal (reinstalls a non-executable current shim; still no-ops on a healthy one).

Full suite: 10267 passed, 24 skipped. `tsc --noEmit` clean, build clean.

## Not covered

The config-path list is the conflict detector's, which is JSON-only — Codex TOML, Hermes YAML and AMP JSONC registrations are not scanned. Those clients are also the ones the conflict scanner skips today; widening that list is one change for both callers, not this one.

Closes TRA-913.

### DIFF
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -16,9 +16,11 @@ import {
   getLauncherConfigPath,
   getLauncherDir,
   getLauncherPath,
+  isBroken,
   readInstalledLauncherVersion,
   readLauncherConfig,
 } from '../init/launcher.js';
+import { checkRegisteredLaunchers, type RegisteredLauncherCheck } from '../init/launcher-health.js';
 import { LAUNCHER_VERSION } from '../init/types.js';
 import { findProjectRoot } from '../project-root.js';
 import {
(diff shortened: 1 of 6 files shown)


nikolai-vysotskyi:

PR #992, opened 2026-09-06
feat(telemetry): privacy page, first-run disclosure notice, config opt-out

nikolai-vysotskyi:
Closes TRA-887.

The anonymous daily ping was documented but never disclosed before it fires, and its only opt-out was an environment variable — the one place a user editing their trace config would not look. Worse, `telemetry.enabled` already exists and means something else (the local latency sink), so setting it to `false` bought a privacy-minded user a reasonable but wrong belief that the ping had stopped.

## What changed

**1. `docs/privacy.md` → https://trace-mcp.com/privacy.html** — in `_data/docs_nav.yml`, `sitemap.xml` and `llms.txt`. The complete field list moved out of README verbatim, plus the public-credentials note, both opt-outs, and how to delete local state (`~/.trace/`). README keeps a five-line summary and links out; `docs/telemetry.md`'s disclaimer now points there instead of at a README anchor; SECURITY.md links it too.

**2. One-time notice on stderr.** `printTelemetryNoticeOnce()` gated on a new `noticeShown` flag in the telemetry state file. Two call sites, both idempotent through that flag:
- a commander `preAction` hook in `src/cli.ts` — `trace init` / `doctor` / `status` are where a human is actually looking at a terminal (skipped for `classify-env`, which the shell guard runs on every tool call);
- server construction in `src/server/server.ts`, the fallback for installs that never touch the CLI.

Silent when telemetry is already off, and silent in a build with no baked-in GA credentials — a source checkout never pings, so the line would be a lie. stderr, not stdout: `armStdoutGuard()` reroutes stdout to keep JSON-RPC framing clean, and stderr survives on both paths.

**3. `telemetry.usage_ping: false`** in `~/.trace/.config.json` disables the ping. Deliberately *not* `telemetry.enabled` — that name is taken. Global config only (install-scoped choice, and the project reader is async while every caller here is sync and best-effort), memoized per process, documented with the distinction spelled out on the privacy page. `docs/config-index.md` regenerated.

Out of scope, as the issue says: no public adoption numbers.

## Test evidence

- `src/telemetry/__tests__/usage-ping.test.ts` +6 tests: config opt-out silences the ping; `telemetry.enabled: false` does *not* (that key is the local sink); notice prints once and never again; it names both opt-outs and the privacy URL; silent when telemetry is off; silent without GA credentials; marking `noticeShown` preserves install id and ping state.
- Full suite: 10256 passed, 1 unrelated flake (`tests/ai/onnx.test.ts` cold-import timeout, green on re-run). `pnpm typecheck`, `pnpm build`, `biome ci` clean. `tests/docs` green including `internal-links`, `page-dates` and `config-index --check`.
- Manual, against a throwaway `TRACE_MCP_DATA_DIR`: first `trace doctor` prints the line, second prints nothing, state file carries `noticeShown: true`; with `{"telemetry":{"usage_ping":false}}` in the config nothing is printed and nothing is sent.


nikolai-vysotskyi:
Merged without a second-model review — both reviewers are out of provider quota.

- Reviewer B: `agent_error.provider_quota_limit`, "Resets in 115h37m" (≈2026-09-10). It has been unavailable since 2026-09-05 (TRA-854), well past the 4-hour degradation threshold.
- Code Reviewer, the fallback, called in the same run: "You've hit your usage limit … try again at 9:04 AM."

Degradation conditions checked before merging: CI green on every check; the diff touches no secrets, no CI workflow, no on-disk graph/DB schema, and no MCP tool signature. The telemetry state file gains one additive optional field (`noticeShown`), which older builds ignore. Author takes responsibility for this merge; it goes back through a reviewer once one is available.

PR #993, opened 2026-09-06
fix(indexer): resolve Astro/Svelte imports, no new pass needed (TRA-451)

nikolai-vysotskyi:
## Summary

Follow-up to TRA-449/#593, which found that a plugin extracting `import` statements is not enough — the specifier has to be resolved to a target node by a pipeline pass, or the edge is dropped. Go, Java, Rust, C/C++, and Ruby have each closed this gap for their language via a dedicated resolver pass. This PR closes it for Astro and Svelte — but neither needed a new resolver.

Both plugins already emit `imports` edges carrying plain filesystem-path specifiers:
- Astro's frontmatter/`<script>` blocks are re-parsed with the same tree-sitter helper TypeScript uses (`extractImportEdges`).
- Svelte's regex plugin already writes `module` specifiers in the shape `resolveEsmImportEdges` expects.

The only gap was that neither language was in `ESM_IMPORT_LANGUAGES`, so `resolveEsmImportEdges` skipped every file of that language outright. Fix:
- Add `astro` and `svelte` to `ESM_IMPORT_LANGUAGES` (`import-capable-languages.ts`) — this also updates the derived `IMPORT_EDGE_LANGUAGES` set and `docs/language-matrix.md`'s Imports column via `pnpm run docs:language-matrix`.
- Add `.astro` to the oxc-resolver extension list in `es-modules.ts` (`.svelte` was already present, for the existing TS→`.svelte` direction).

Of the original 13 languages in TRA-451, six are now done (Go, Java, Rust, C, C++, Ruby, plus Astro/Svelte = 8); five remain unresolved: C#, Kotlin, Swift, Elixir, Lua.

## Test plan

- [x] New e2e test `tests/integration/astro-svelte-import-resolution-e2e.test.ts` against an on-disk fixture (`tests/fixtures/astro-svelte-imports/`), covering: Svelte→Svelte, Svelte→TS, Astro frontmatter→Astro, Astro frontmatter→TS, and the pre-existing TS→Svelte/Astro direction. All 5 pass.
  - Uses an on-disk fixture rather than `createTmpFixture`: `resolveEsmImportEdges` resolves through oxc-resolver, which returns realpath'd targets, and `os.tmpdir()` is a symlink on this sandbox — the realpath'd target then falls outside `state.rootPath` and the edge is wrongly treated as external. Matches the existing pattern in `split-imports.test.ts`.
- [x] `pnpm run build` — passes.
- [x] `pnpm run typecheck` — passes.
- [x] `npx biome ci` on all changed files — clean.
- [x] Full suite: `npx vitest run` → 10250 passed, 4 failed, 26 skipped. The 4 failures (`locate-app.test.ts`, `postinstall-app.test.ts`) are pre-existing sandbox flakiness in unrelated subsystems (real `pgrep`/`mdfind` subprocess timing) — confirmed by re-running them in isolation both with and without this diff; failures are non-deterministic and occur in either state.
- [x] `pnpm run docs:language-matrix` regenerated `docs/language-matrix.md`: import edges 16 → 18, astro/svelte rows flip to `yes`.

### DIFF
--- a/docs/language-matrix.md
+++ b/docs/language-matrix.md
@@ -48,7 +48,7 @@ is the two-pass pipeline in [architecture](architecture.md#indexing-pipeline).
 
 - **indexed with the default config:** 78 (the rest need an `include` entry — see below)
 - **tree-sitter parser:** 29 · **regex parser:** 46 · **custom parser:** 6
-- **import edges:** 16
+- **import edges:** 18
 - **call edges (`get_call_graph`, call-aware `find_usages`):** 3
 - **type / inheritance edges:** 2
 - **covered by a plugin test:** 68
@@ -100,7 +100,7 @@ second thing, so it is much shorter than the language list — see
 | al | .al | regex | yes | — | — | — | yes |
 | apex | .cls .trigger .apex | regex (multi-pass) | yes | — | — | — | yes |
 | assembly | .asm .s .S | regex | yes | — | — | — | yes |
-| astro | .astro | tree-sitter | yes | — | — | — | yes |
+| astro | .astro | tree-sitter | yes | yes | — | — | yes |
 | autohotkey | .ahk .ah2 | regex | yes | — | — | — | yes |
 | bash | .sh .bash .zsh | tree-sitter | yes | — | — | — | yes |
 | blade | .blade.php | regex | yes | — | — | — | yes |
@@ -163,7 +163,7 @@ second thing, so it is much shorter than the language list — see
 | scala | .scala .sc | tree-sitter | yes | — | — | — | yes |
 | solidity | .sol | tree-sitter | yes | — | — | — | yes |
 | sql | .sql | regex | yes | — | — | — | yes |
-| svelte | .svelte | regex | yes | — | — | — | yes |
+| svelte | .svelte | regex | yes | yes | — | — | yes |
 | swift | .swift | tree-sitter | yes | — | — | — | yes |
 | tcl | .tcl .tk .itcl .itk | regex (multi-pass) | yes | — | — | — | yes |
 | toml | .toml | tree-sitter | yes | — | — | — | yes |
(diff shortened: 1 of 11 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Resolves cross-file imports for Astro/Svelte (TRA-451) by adding both languages to `ESM_IMPORT_LANGUAGES` — no new resolver pass needed, since both plugins already emit `imports` edges with plain filesystem-path specifiers in the shape `resolveEsmImportEdges` expects; the gap was purely that the source-language filter skipped them. `.astro` is also added to the oxc-resolver extension list so `.ts → .astro` resolution works (`.svelte` was already there).

### What I checked
- This is exactly the kind of tree-sitter/resolver change that needs a real-language fixture, not a synthetic one — and it has one: `tests/fixtures/astro-svelte-imports/` is a real mini Astro+Svelte+TS project (frontmatter imports, `<script>` imports, a `.ts` file importing both component types), not hand-rolled minimal snippets.
- The new E2E test (`tests/integration/astro-svelte-import-resolution-e2e.test.ts`) indexes the fixture through the real pipeline and asserts on actual graph edges (`store.getOutgoingEdges` → `imports` edge type), not on resolver internals — it would catch a real regression, not just a unit-level illusion of coverage. It also pins the pre-existing reverse direction (`.ts` importing `.svelte`/`.astro`) so the PR can't silently break what already worked while fixing what didn't.
- Confirmed `ESM_IMPORT_LANGUAGES` has exactly one consumer (`resolveEsmImportEdges` in `src/indexer/edge-resolvers/imports.ts:113`), so adding two entries has no other blast radius to check.
- `docs/language-matrix.md`'s count bump (16→18 import-edge languages) matches the two languages added, and `tests/docs/language-matrix.test.ts` (a live-count test, not hand-typed) passes.

### Findings
None.

### What I ran
- `npx vitest run tests/integration/astro-svelte-import-resolution-e2e.test.ts` → 5 passed (6 edges resolved, 0 phantom)
- `npx vitest run tests/integration/split-imports.test.ts tests/resolvers/es-modules.test.ts tests/docs/language-matrix.test.ts` → 8 passed
- `npx vitest run tests/languages/svelte.test.ts` → 7 passed

Merging.

Issue #957, opened 2026-09-05
search() returns 'Maximum call stack size exceeded' on large indexes (~150k symbols); small-index fallback path works


msalem89:
**Environment**
- trace-mcp v3.18.0 (installed via `npm install -g trace-mcp`), Node v22.22.3, macOS 26.6.2 arm64
- Server: `trace serve` (stdio), cwd = project root
- Index scale: **10,332 files / 152,734 symbols** (per the `evidence` block of a working `search_text` call). Registered via `trace-mcp add .` — 9,963 files indexed in 77.5s, 37 errors, 0 skipped.

**Repro**
1. Index a project of ~10k files
2. Start `trace serve` with cwd set to the project root
3. MCP `initialize` → `tools/call` name=`search`, arguments `{"query":"decision_proxy","limit":5}`

**Actual:** the tool returns HTTP-200-style success with content text exactly:

```
Maximum call stack size exceeded
```

(not JSON, not an MCP error envelope — the string itself is the tool result)

**Expected:** a JSON result with `items`, or a structured error.

**Data points**
- 3/3 deterministic across two different query terms (`decision_proxy`, `authorize`)
- `search_text` on the same index works fine and returns real hits, so the index itself is readable
- On a small repo (~100 files) `search` works — but via `"fallback":true, "reason":"Using on-the-fly search — index unavailable or stale"`, i.e. the small-index path never reaches the crashing code path
- Suspect: recursion or unbounded recursion depth in the indexed symbol-search path that only triggers above some index-size threshold

**Secondary observation (possibly related):** `get_change_impact` with `{"file_path":"tools/decision_proxy.py"}` — a path that exists relative to the project root and is provably in the index (search_text hits it) — returns:

```json
{"error":{"code":"NOT_FOUND","message":"'' is not in the index"}}
```

The empty `''` in the message suggests the `file_path` field was not picked up despite matching the published input schema.

**Workaround:** use `search_text` (content search) and `get_outline`, then impact via `symbol_id`.

Happy to run additional diagnostics or test a fix branch.

nikolai-vysotskyi:
Reproduced and fixed — thank you, this was an excellent report. The detail that made it fast was yours: `search_text` working on the same index while `search` failed, plus the note that the small repo went through `"fallback": true`. That pair ruled out the index and pointed straight at a code path only large indexes enter.

[...]

PR #997, opened 2026-09-06
test(scale): run the real tools against a 150k-symbol index (TRA-977)

nikolai-vysotskyi:
Closes TRA-977. Builds on #976 (contains its commit; will be a clean diff once that merges).

## What this is

#976 fixed two `Math.min(...)` call sites. This covers the **class**: a defect that only exists above ~65k elements, against a corpus that never reached it.

## The number the issue asked for

Largest corpus CI ran: **50 000 symbols** (`tests/perf/stress.test.ts`, 10 000 files × 5) — an order below V8's argument limit, and it only exercised `searchFts` and store methods, never a tool handler.

## The audit turned the diagnosis around

`searchFts` applies `LIMIT ?` **in SQL**. Every `ftsResults` array is therefore bounded by the caller's `limit` (~70–120 rows) and cannot reach 65k. Those spreads were never the crash.

The array that scales with the index is `pagerankMap` — one entry per graph node — spread as `Math.max(...pagerankMap.values(), 0.001)` in three places (`search()`, the fusion path, `get_feature_context`). **All three still threw on a 150k index after #976's fix.** The new test reproduces each by name:

```
RangeError: Maximum call stack size exceeded
 ❯ search src/tools/navigation/navigation.ts:433:22
 ❯ getFeatureContext src/tools/navigation/context.ts:208:22
```

`task-context.ts` had already been written as a manual loop — the one correct precedent in the tree.

## What landed

| | |
|---|---|
| `tests/perf/large-index.ts` | The `stress.test.ts` seeder, extracted, plus symbol-level edges. Generated, never vendored. PageRank ranks only nodes in a resolved edge, so a file-only edge set caps `pagerankMap` at the file count and cannot reproduce the failure at all. |
| `tests/perf/scale-rangeerror.test.ts` | 150 000 symbols / 165 000 nodes, real tool entry points, **~4 s**. Sized above the *upper* end of the V8 range — a 70k fixture passes on a deeper stack and greens a crash the user still gets. |
| `tests/ci/no-spread-into-call.test.ts` | Total ban on `Math.min/max(...)` in `src/**`, **no allowlist** — `minMax()` is a drop-in for every form, so there is no legitimate exception and nothing to maintain. Verified to fail on reintroduction. |
| 26 call sites | Converted to `minMax()`. `minMax` widened to `Iterable<number>` so `pagerankMap.values()` reduces without materialising a second 150k copy. |
| `ops/index-scale-testing.md` | The threshold, the audit, and what is deliberately **not** gated. |

## Deliberately not gated

`arr.push(...xs)` throws the same `RangeError`, ~40 sites, most bounded. The safe rewrite is not a drop-in, and a rule with 35 exemptions is an allowlist wearing a rule's clothes. The scale test covers the paths it exercises; recorded as standing risk. Revisit on a field report, not on a guess.

## Test evidence

`pnpm test` — 925 files / 10 248 passed. One unrelated pre-existing flake (`tests/ai/onnx.test.ts`, cold model import under full-suite contention; passes in isolation). `pnpm typecheck` clean, `biome check` clean.

Original report and repro: @msalem89 in #957. Cause and the first fix: Nikolai, in that thread.

### DIFF
--- a/ops/index-scale-testing.md
+++ b/ops/index-scale-testing.md
@@ -1,0 +1,100 @@
+# Index-scale testing: the ceiling we test to, and why that number
+
+A defect that only exists above a size threshold is invisible to a corpus below
+it. This file records where our threshold sits, so the next run does not
+re-derive it — and so that a fixture shrunk in a refactor is a visible change
+rather than a silent loss of coverage.
+
+## The threshold that matters
+
+`Math.min(...xs)` and `arr.push(...xs)` pass every element as a separate
+argument. V8 throws `RangeError: Maximum call stack size exceeded` past its
+argument limit — **~65 000 to ~125 000, depending on remaining stack depth**.
+Not a fixed number: the same array can pass on one machine and throw on
+another, and can pass in a shallow call and throw in a deep one.
+
+Practical consequence: a fixture at 70k proves nothing. A fixture has to clear
+the *upper* end of that range to be evidence.
+
+## Where we were (2026-09-06, before this change)
+
+| Corpus | Symbols | What it exercised |
+|--------|--------:|-------------------|
+| `tests/perf/stress.test.ts` | **50 000** (10 000 files x 5) | `searchFts`, store batch methods |
+| everything else | < 1 000 | tools, end to end |
+
+So the largest thing CI ran was **50 000 symbols** — an order below the
+threshold — and even that only reached the DB layer, never a tool handler. The
+size the defect needs and the surface it lives on were both outside CI.
+
+That is how GitHub [#957](https://github.com/nikolai-vysotskyi/trace-mcp/issues/957)
+reached a user: `search` returned the bare string "Maximum call stack size
+exceeded" for every query against a 152 734-symbol index, deterministically,
+while `search_text` on the same index worked. Reported by `msalem89` with a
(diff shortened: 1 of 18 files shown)


nikolai-vysotskyi:
Merging with the second-model gate in **degradation mode** — recording that explicitly, per the workspace rule, because no reviewer saw this.

Both reviewers are quota-blocked with published reset times, so the outage is known to exceed 4 hours rather than being a wait I could sit out:

- **Reviewer B** (primary) — `Individual quota reached. Resets in 115h38m` (dispatched 00:20 UTC, failed 00:22).
- **Code Reviewer** (fallback) — `You've hit your usage limit... try again at 9:04 AM` (dispatched 00:43, failed 00:43).

The diff clears all four protected areas the gate exists for: no CI workflow files, no on-disk schema change, no MCP tool signature change (only how `maxPr` is computed inside three handlers — parameters and return shapes are untouched), no secrets. CI is 20/20 green.

I am personally responsible for this merge. When Code Reviewer is back this afternoon, this PR should be run through it and anything found fixed in a follow-up — the two things worth their attention are the claim that `searchFts`'s SQL `LIMIT` makes every `ftsResults` array bounded (which is what makes #976's annotation wrong about its own line), and whether 150 000 is comfortably above the worst-case V8 argument limit on CI runners.

PR #998, opened 2026-09-06
docs(site): give door 2 a page — what `trace init` installs (TRA-919)

nikolai-vysotskyi:
Closes TRA-919.

## The gap

The product ships through two doors, and the unit that decides a door is **how a thing is installed**, not which mechanism it belongs to.

- **Door 1 — the tool surface.** Graph, decision memory, `trace_state_*`, `get_startup_context_audit`. Installed by adding an MCP server, governed by presets. Well documented.
- **Door 2 — what `trace init` installs.** The hooks, the guard, the startup-audit apply button, the config edits. `hooks/` holds 23 scripts, and **none of it had a name on any public surface.** A reader who installs us as "an MCP server" never learns that half of what they installed exists.

## What's in here

**One new page** — `/what-trace-init-installs.html`.

It **leads with the startup-block audit**, not the mirrors. `get_startup_context_audit` measures the reader's *own* block, which sidesteps the "one machine, one client, one model" limit that applies to every figure in `ops/context-block-levers.md`.

The mirrors are reported as what they actually measured: 52% of the paid band compressed offline, **0 pp of solve-rate change across 108 live runs**, opt-in behind `--mirror`. Cost down, capability flat — in those words, no dressing.

Every `tweakcc` mention draws the distinction explicitly: **it is a third-party tool the user installs and runs, we document a pairing, trace-mcp patches nothing.** The page also states what door 2 does *not* reach — the guard and the pairing are Claude Code / Claw only, Cursor and Windsurf get a rules file, and nothing here has been measured on another client.

**A two-card `.doors` band on the landing**, one card per door — door 1 points at the existing tools reference, door 2 at the new page. Two cards and not four: splitting per-mechanism re-creates the "two products in one binary" reading the positioning pass exists to avoid.

## Where it sits, and why

Below the metrics strip, not above it. §8's first screen is a measurement, and this block is not first-screen material — it is where a reader goes after the evidence. Measured after the change: `.hero` 48→827, `.metrics-strip` 827→1294, `.doors` 1294→1813 — the two bands above it occupy exactly what they did before.

## Measured, not eyeballed

Headless Chrome, isolated profile, both themes:

- 1440×900 dark and light — both `.door-go` labels bottom at **575px** inside cards bottoming at **608px**. The flex column with `margin-bottom: auto` on the last paragraph is what buys that; without it the shorter card's link floats up mid-card and the pair reads as two unrelated blocks.
- 390×844 — one column, 350px cards, `scrollWidth == clientWidth == 390`, no horizontal overflow.
- Monochrome throughout, square corners, hover raises the outline to `--text-display` and that is the only state (same treatment as `.app-frame`). No accent fill: the accent is spent on actions, and a door is a link. No red anywhere.
- Focus ring comes free — `.door` is an `<a>`, covered by the global `:where(a, button, …):focus-visible` rule.

`docs/DESIGN-WEB.md` §11 records the component and both measurements.

## Tests

`tests/docs` — 250 passed, 4 skipped. Full suite: 10,247 passed, 4 failed, all four in `tests/scripts/postinstall-app.test.ts`, which passes 17/17 when run on its own — process-spawning flake under parallel load, and the diff here touches only `docs/`.

## Note for the SEO agent

The page is registered in `sitemap.xml`, `docs_nav.yml`, `llms.txt` and the regenerated `llms-full.txt`. Nav membership is yours — this is a new page needing a home, not a reordering, but the title and the position in the list are open to your call.

Agent: Web Design Agent


nikolai-vysotskyi:
**Review gate status — not merged, and here is why.**

Reviewer B was unavailable (provider quota, ~115h to reset), so the gate fell to Code Reviewer, who returned **NEEDS WORK** with six findings. All six are fixed in `f78e63e4` — the load-bearing one being that the page described a `tweakcc` flow the code does not run: non-interactive `trace init` defaults Claude-family clients to Max, Max calls `installTweakccPrompts`, and that shells out to `npx tweakcc --apply` itself. The page now says that plainly, and the wider problem — `ops/positioning.md` obliges every public surface to repeat "trace-mcp does not patch anything" — is filed as TRA-979.

The re-review task then failed on quota as well, so **both reviewers are currently unavailable and this PR has no APPROVE.** The degraded-merge path needs both down for four hours straight; that has not elapsed. Leaving it open rather than merging on my own reading of my own fixes.

Next run, or whoever picks this up: re-request review, and re-check findings 1 and 2 in particular — both are claims about code paths, not wording.

nikolai-vysotskyi:
## Verdict: APPROVE

New docs page (`what-trace-init-installs.md`) plus nav/sitemap/llms.txt wiring and a DESIGN-WEB.md rationale section for the homepage's new "two doors" IA block. No product code touched.

### What I checked
- This page makes a lot of precise, checkable claims (hook thresholds, percentages, env var names, tool names), so I verified the load-bearing ones against the actual implementation instead of trusting the prose:
  - `get_startup_context_audit`, `apply_startup_recommendations`, `rollback_startup_recommendations` — all real, found in `src/analytics/startup-watch.ts` / `apply-recommendations.ts` / `src/tools/register/session.ts`.
  - Guard hook: page says "redirected from the third attempt inside a five-minute window" via `TRACE_MCP_GUARD_NAV_MIN` / `TRACE_MCP_GUARD_NAV_WINDOW` — `hooks/trace-mcp-guard.sh:481-482` defaults `NAV_STREAK_MIN=3`, `NAV_STREAK_WINDOW_SEC=300`. Matches exactly.
  - Mirror hook: page says "24/12-line window... 52% offline... 62.6% wider cap... `TRACE_MCP_MIRROR_CAP`... `TRACE_MCP_MIRROR_DISABLE`" — `hooks/trace-mcp-mirror.sh:29-59` states the same 24/12 defaults, "52.0% offline with 0pp solve-rate loss," and the same two env var names. The 52%/62.6%/108-run figures aren't new to this PR — they're already in `docs/ROADMAP.md` and `ops/positioning.md`, so this page is restating existing measurements, not introducing unverified ones.
  - `tweakcc` pairing — `src/init/tweakcc.ts` exists and matches the described `~/.tweakcc/system-prompts/` + `npx tweakcc --apply` mechanism.
- Ran the full `tests/docs/` suite, including `internal-links.test.ts` (broken-link / nav-consistency checks) and the numeric-claims suites — **254 passed, 4 skipped, 0 failed**.

### Findings
None.

### What I ran
- `npx vitest run tests/docs/` → 254 passed, 4 skipped

Merging.

PR #999, opened 2026-09-06
fix(indexer): scale ExtractPool to zero within 60s of daemon idle (TRA-971)

nikolai-vysotskyi:
## Summary

- `KEEPALIVE_IDLE_TERMINATE_MS` (`src/indexer/extract-pool.ts`) dropped from 5 minutes to 45 seconds, so a daemon that isn't indexing gives back its ~300 MB of worker-thread RSS well inside the 60s idle bar instead of staying warm for most of a workday.
- Added a per-instance `keepAliveIdleMs` override on `ExtractPoolOptions`, mirroring the existing `workerEntry` override pattern, so a test can assert on scale-to-zero without waiting out the real default.
- Updated the in-code rationale comment and `docs/perf/daemon-idle-memory.md`'s TRA-811 reading to record the follow-up.

Scale-to-zero itself (idle teardown, lazy on-demand respawn, DI-shared pool across pipelines) already existed from TRA-811/TRA-925/TRA-931 — this closes the remaining gap, the hardcoded 5-minute window, called out in TRA-971.

## Test plan
- [x] `pnpm run test` — full suite green (10249 passed, 26 skipped)
- [x] Targeted: `tests/indexer/extract-pool-keepalive.test.ts`, `tests/daemon/shared-extract-pool.test.ts`, `tests/perf/extract-pool-lazy-spawn.test.ts` — all pass
- [x] `pnpm run build` (tsup) — clean
- [x] `pnpm run lint` (tsc --noEmit) — clean

### DIFF
--- a/docs/perf/daemon-idle-memory.md
+++ b/docs/perf/daemon-idle-memory.md
@@ -74,3 +74,9 @@ Expected steady-state effect on an idle daemon with 8 workers: about **-420 MB**
 taking the idle median from ~950 MB toward ~530 MB. Not yet confirmed in the field; the
 daemon on the measuring machine restarts every ~3 min (TRA-809), so it never sits idle
 long enough to cross the 5-minute window there.
+
+**TRA-971 follow-up:** 5 minutes was still long enough that a daemon idle for under an
+hour of a workday kept the pool warm the whole time. `KEEPALIVE_IDLE_TERMINATE_MS` is now
+45 seconds — comfortably inside the 60s idle bar from the acceptance criteria — with a
+per-instance `keepAliveIdleMs` override on `ExtractPoolOptions` so tests don't have to
+wait out the real default.
--- a/src/indexer/extract-pool.ts
+++ b/src/indexer/extract-pool.ts
@@ -7,9 +7,9 @@
  * files actually proceeds in parallel.
  *
  * Lifecycle: lazy. The first `extract()` call spawns workers; they are released
- * again after an idle window (milliseconds for CLI, minutes for the daemon —
- * see IDLE_TERMINATE_MS / KEEPALIVE_IDLE_TERMINATE_MS) and `terminate()` shuts
- * them down for good. Worker spawn cost is non-trivial (~150-300 ms each
+ * again after an idle window (milliseconds for CLI, tens of seconds for the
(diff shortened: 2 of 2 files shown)


nikolai-vysotskyi:
@Reviewer B — review requested.

Small, focused diff on `src/indexer/extract-pool.ts` (+ a docs/perf ledger note): drops `KEEPALIVE_IDLE_TERMINATE_MS` from 5 min to 45s and adds a `keepAliveIdleMs` constructor override for tests. No API/behavior change beyond the constant value and the new optional field. Please look especially at whether the `keepAliveIdleMs ?? KEEPALIVE_IDLE_TERMINATE_MS` wiring in the constructor and `scheduleIdleTeardown` is correct, and whether 45s is a reasonable choice given the 60s acceptance bar in TRA-971.

nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Shortens `ExtractPool`'s keepAlive idle-teardown window from 5 minutes to 45 seconds (TRA-971's 60s acceptance bar), and adds a per-instance `keepAliveIdleMs` override so tests don't have to sleep 45s for real.

### What I checked
- Confirmed the override is correctly scoped: `this.keepAliveIdleMs = o.keepAliveIdleMs ?? KEEPALIVE_IDLE_TERMINATE_MS` is set unconditionally in the constructor, but only ever consumed via `this.keepAlive ? this.keepAliveIdleMs : IDLE_TERMINATE_MS` — so the doc comment's "ignored when `keepAlive` is false" is accurate; a non-keepAlive (CLI/test) pool still uses the untouched 200ms `IDLE_TERMINATE_MS`.
- Cross-checked the updated doc comment's "pool of 4" claim against `DEFAULT_KEEPALIVE_WORKER_COUNT = Math.max(1, Math.min(4, cpus/2))` — accurate (this constant wasn't touched by this PR, the comment is just catching up to it).
- Grepped for other spots hardcoding the old 5-minute constant (`300_000` / `5 * 60_000`) — none found, so nothing else silently assumes the old window.

### Findings
None.

### What I ran
- `npx vitest run tests/indexer/extract-pool-keepalive.test.ts` → 7 passed

Merging.

PR #1000, opened 2026-09-06
fix(app): every Clients row names its own state (TRA-479)

nikolai-vysotskyi:
Closes TRA-479.

The parked lead needed a fixture with a **connected, current** client in it before the grouping could be judged honestly. I built one (scripted `getMcpClientStatuses` + scripted daemon, renderer driven in headless Chrome) and rendered all six states side by side. The conclusion the fixture produced: the states *are* genuinely per-client, so no new grouping — what was missing is that four of the six states did not name themselves.

### 1. A `missing` row said nothing

`detectClientStatus` folds two facts into `missing`: a config file that exists with no trace-mcp entry in it (one click from working) and no config file at all (usually the client isn't installed). Both rendered as a bare `Windsurf · Connect` — the only rows on the screen with no caption, while every row above them carried a path.

`McpClientStatus` now carries `configExists`, and the row says which: the file `Connect` will write into, or `No config file found`. A daemon older than the field reports neither and the row claims neither.

### 2. Six identical badges for one condition

`Update available` badge + `Update` button + a header already counting them = one condition said three ways (`DESIGN.md` §5), and after any upgrade it's every configured client at once. The badge goes; the verb names the state and the next step together, and the drifted field moves onto the button's `title`. Same for `legacy`/`Migrate`.

### 3. `unknown` was impersonating `up_to_date`

Codex is TOML — we detect the entry and cannot compare it. It rendered as the same green dot and the same word `Connected` as a config we verified. Now a neutral dot and `Configured`, with the reason on hover.

### 4. Off-grid row height

Caption rows measured 46px (padding, not height, was setting it). Since almost every row now has a caption, that set the list's rhythm. Rows are 44 bare, 48 with a caption — both on the 4pt grid.

### Verification

Rendered light and dark on the running renderer, before and after; screenshots on the issue. Measured on the running page: caption `rgba(0,0,0,.55)` on `#fff` = **4.74:1**, `rgba(255,255,255,.55)` on `#1e1e1e` = **5.36:1** — both pass 4.5:1. Row heights read back 44/48. Reduce Transparency is not involved: this surface's card is opaque content, and the only glass (the toolbar) is untouched.

`pnpm test` green (714 app + root suites), typecheck green both packages, `check:i18n` clean, both builds pass. New tests cover the config-file split, the unknown-daemon case, the absent badge, and presence-only detection. Ten locales updated; `updateAvailable` / `legacyEntry` removed everywhere.

The two rules this produced are written into `DESIGN.md` §5 so the next screen gets them for free.

Agent: Design/UX Agent


nikolai-vysotskyi:
## APPROVE — PR #1000 (TRA-479)

Checked out `tra-479-clients-row-states` (commits `083c3ec9` + `3026208b`), read every file in the diff via `git diff master...`, ran the actual test/build commands rather than trusting the description.

### What I checked and what I found

**`src/init/mcp-client.ts` / non-interactive path (worry #1 in the brief).** `configExists` is a single new optional field, added at the one call site inside `getMcpClientStatuses` (`...status, level, configExists`). `detectClientStatus`'s return type is typed `Omit<McpClientStatus, 'level' | 'configExists'>`, so the compiler forces every code path to set it — no way to silently forget it on a future branch. Ran `pnpm run test --run tests/init/mcp-client-status.test.ts`: 28/28 pass. Built the CLI (`pnpm run build`) and ran `node dist/cli.js clients status --json --scope global` for real against my own machine's configs — output is the old shape plus one trailing key, nothing renamed or removed. This is additive, not a breaking contract change; `trace init`'s non-interactive path doesn't branch on this field at all.

**preload → renderer bridge (worry #2).** `configExists?: boolean` was added in lockstep to `preload.ts` and `electron.d.ts` (both optional, matching each other exactly), and threaded through to `<SupportedClientRow>` in `Clients.tsx:748`. No renderer code assumes it's always present — an older daemon that doesn't send it renders as `undefined`, and the row correctly falls back to "say nothing" (covered by its own test, see below). No `undefined`-where-string-expected risk.

**i18n catalogs (worry #3).** Diffed all 10 languages. Every one adds exactly `configured`, `configuredHint`, `noConfigFile` and removes exactly `updateAvailable`, `legacyEntry` — same three additions, same two removals, in every catalog, all genuinely translated (checked Hindi specifically per TRA-804: the three new strings are in Devanagari, not English placeholders). Grepped for leftover references to the two removed keys (`updateAvailable`, `legacyEntry`) anywhere under `packages/app/src` — none. `catalog-parity.test.ts` ran as part of the full suite and passed.

**Accessibility (worry #4).** Connected/Configured states are dot-tone *and* word (`t('connected')` / `t('configured')`), never color alone. Missing/Update/Migrate are distinguished by button label text, not color. `StatusDot`'s `tone="neutral"` is a real, pre-existing tone (used elsewhere in the codebase), not invented for this PR.

**Removed `Badge` import.** Confirmed zero remaining `Badge` usages in `Clients.tsx` after the edit — not a dead import.

### Tests actually run
- `pnpm run test --run tests/init/mcp-client-status.test.ts` → 28/28 pass
- `pnpm run test` (packages/app) → 70 files, 714/714 pass (includes `Clients.test.tsx`, both `catalog-parity.test.ts` files, `i18n.test.ts`)
- `pnpm run typecheck` (packages/app) → clean
- `pnpm run build` (root, tsup) → clean

### Findings
None. Looked specifically for: a JSON contract break in `clients status --json` (none — purely additive), a preload/renderer field mismatch (none — added to both in the same commit, both optional), an i18n key present in `en` but missing elsewhere (none — checked all 10 catalogs by hand), a color-only state signal (none — every state pairs a dot with a word or a distinct button label), and a stale `Badge` import (none).

One cosmetic nit not worth blocking on: `DESIGN.md` picks up a double blank line before the closing `---` (line ~820). Not touching it.

Not merging — per my role, merge is Lead Engineer's call.

nikolai-vysotskyi:
## Verdict: APPROVE

Two UI fixes to the desktop app's Clients screen (TRA-479): (1) removes redundant `Badge` + `Button` pairs that said the same state twice ("Update available" badge next to an "Update" button, "Legacy" badge next to "Migrate") — the reason moves to the button's `title` tooltip instead; (2) adds `configExists` end-to-end (`src/init/mcp-client.ts` → CLI `clients status --json` → Electron IPC → renderer) so a `missing` row can distinguish "config file exists, just no trace-mcp entry" from "no config file at all," instead of rendering both as the same bare "Connect" row.

### What I checked
- Traced the `configExists` plumbing end-to-end since it crosses a process boundary I couldn't just grep once: `src/init/mcp-client.ts` sets it as a required `boolean` on `McpClientStatus`; `src/cli/clients.ts`'s `clients status --json` serializes the whole array via `JSON.stringify({ scope, projectRoot, statuses })` (no allowlist, so the new field isn't silently dropped); `packages/app/src/main/index.ts`'s IPC handler forwards `parsed.statuses` from the CLI's raw JSON.parse output untouched (its local TS type annotation wasn't updated to name `configExists`, but that's just an unchecked type hint on an `any`-shaped JSON blob — the value still flows through at runtime); `preload.ts` and `electron.d.ts` type it `configExists?:` (optional, matching the doc comment that an older daemon simply won't report it, which the renderer treats as "say nothing" rather than guessing).
- Confirmed `connected` narrowed from `status === 'up_to_date' || status === 'unknown'` to just `'up_to_date'`, with `'unknown'` (Codex TOML, presence-only) now a distinct `presenceOnly` state rendered as a neutral dot + "Configured" — correctly stops claiming a drift-comparison that was never run.
- Local environment note, not a PR defect: `packages/app` is its own standalone package (own `pnpm-lock.yaml`, not part of a root pnpm workspace) and its `node_modules` was never installed in this checkout, so `Clients.test.tsx` initially failed on `react/jsx-dev-runtime` resolution — reproduced identically on master's copy of the same test, confirming it was a local setup gap, not something this PR broke. Ran `pnpm install` inside `packages/app/` to fix it, then re-ran clean.
- `npx tsc --noEmit` (both root and `packages/app`) — clean.

### Findings
None. (Minor, non-blocking: `packages/app/src/main/index.ts`'s local return-type annotation on the `get-mcp-client-statuses` IPC handler wasn't extended with `configExists`, unlike `preload.ts`/`electron.d.ts`. Cosmetic type-completeness gap only — the JSON value isn't type-checked against it and passes through fine.)

### What I ran
- `npx vitest run tests/init/mcp-client-status.test.ts` (root) → 28 passed
- `cd packages/app && npx vitest run src/renderer/tabs/__tests__/Clients.test.tsx` → 20 passed
- `npx tsc --noEmit -p tsconfig.json` (root and `packages/app`) → clean

Merging.

PR #1001, opened 2026-09-06
perf(daemon): serve first requests from the index snapshot before daemon handshake (TRA-948)

nikolai-vysotskyi:
## Summary
- Adds `SnapshotBackend`: opens the daemon's canonical index DB (`sharedDbPath`) directly — readonly, no seed copy, no DDL/migrations — and serves a full MCP surface from it.
- `StdioSession.bootstrap()` now uses it as the initial backend whenever that DB file already exists on disk, so `stdio.start()` (and thus `initialize` / first tool calls) no longer waits on `watcher.start()`'s daemon `/health` check or `ProxyBackend.start()`'s HTTP registration + SSE handshake (measured 2.7–3.6s cold, TRA-931).
- The real backend (proxy or local) is negotiated off the critical path in the new `settleRealBackend()` and handed off via the existing `MessageRouter.swap()` — draining in-flight requests, replaying the cached `initialize` into the new backend, and closing the snapshot's readonly connection. Candidate backends are passed to `swapTo()` unstarted (matching the existing `onDaemonStateChange`/`fallbackToLocal` convention), so `router.swap()` is the only thing that starts them.
- No shared DB yet (fresh project, never indexed) → `buildSnapshotBackend()` returns `null` and `bootstrap()` falls through to the exact pre-existing flow, unchanged.

## Test plan
- New: `src/daemon/router/__tests__/snapshot-backend.test.ts` — `SnapshotBackend` answers `initialize`/`tools/call get_project_map` from a prebuilt DB; rejects `start()` when the file doesn't exist; never mutates or locks the shared file (`stop()` leaves it openable by a second instance).
- New: `tests/daemon/session-snapshot-fast-path.test.ts` — full `StdioSession` bootstrap against a real snapshot DB and a black-hole "daemon" (accepts TCP, never answers — the same worst case TRA-704's tests use): `initialize` + a `get_project_map` tool call both answer in well under the daemon's 500ms `/health` timeout, then the session hands off to the (mocked) real backend exactly once; the shared DB file survives session shutdown.
- Existing `tests/daemon/session-cold-start-handshake.test.ts` (TRA-704, uses a nonexistent `sharedDbPath`) passes unchanged — confirms the no-snapshot fallback path is behavior-preserving.
- `pnpm run build`, `pnpm run lint` (`tsc --noEmit`), `pnpm run format:check` (biome) all clean.
- `pnpm run test` (full suite): 928 test files / 10,256 tests passed, 24 skipped (pre-existing), exit 0.

### DIFF
--- a/src/daemon/router/__tests__/snapshot-backend.test.ts
+++ b/src/daemon/router/__tests__/snapshot-backend.test.ts
@@ -1,0 +1,105 @@
+/**
+ * TRA-948: SnapshotBackend opens the daemon's canonical index DB directly,
+ * readonly, with no seed copy — unlike LocalBackend, which derives an owned
+ * session temp DB it deletes on stop(). These tests guard the property that
+ * makes the fast path safe: the shared file is never mutated or deleted.
+ */
+import { mkdtempSync, rmSync, existsSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import type { JSONRPCMessage } from '@modelcontextprotocol/sdk/types.js';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { TraceMcpConfigSchema } from '../../../config.js';
+import { initializeDatabase } from '../../../db/schema.js';
+import { SnapshotBackend } from '../snapshot-backend.js';
+
+let tmpDir: string;
+let dbPath: string;
+let backend: SnapshotBackend | null = null;
+
+beforeEach(() => {
+  tmpDir = mkdtempSync(join(tmpdir(), 'trace-mcp-snapshot-backend-'));
+  dbPath = join(tmpDir, 'shared.db');
+  // Pre-create the index DB the way the daemon would — the backend must
+  // never run DDL/migrations on it itself.
+  initializeDatabase(dbPath).close();
+});
+
+afterEach(async () => {
+  await backend?.stop();
+  backend = null;
+  rmSync(tmpDir, { recursive: true, force: true });
+});
+
+/** Drives one request/response round trip over the backend's in-memory transport. */
+function sendAndWait(b: SnapshotBackend, msg: JSONRPCMessage): Promise<Record<string, unknown>> {
+  return new Promise((resolve) => {
+    b.onmessage = (m) => resolve(m as unknown as Record<string, unknown>);
+    void b.send(msg);
+  });
+}
+
+describe('SnapshotBackend (TRA-948)', () => {
+  it('answers initialize and get_project_map from the snapshot without mutating it', async () => {
+    backend = new SnapshotBackend({
+      projectRoot: tmpDir,
+      config: TraceMcpConfigSchema.parse({}),
+      sharedDbPath: dbPath,
+    });
+    await backend.start();
+
+    const init = await sendAndWait(backend, {
+      jsonrpc: '2.0',
+      id: 1,
+      method: 'initialize',
+      params: {
+        protocolVersion: '2024-11-05',
+        capabilities: {},
+        clientInfo: { name: 'test', version: '0' },
+      },
+    } as unknown as JSONRPCMessage);
+    expect(init.error).toBeUndefined();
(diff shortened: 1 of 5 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: NEEDS WORK

`SnapshotBackend` is a clean addition on top of solid, pre-existing infrastructure (`MessageRouter.swap()` already detaches the old backend's `onmessage` before stopping it, and drains/synthesizes errors for in-flight requests; `cachedInitialize` capture already happens upstream of any backend, so replay into whichever backend `settleRealBackend()` builds "just works"). The race the design is most exposed to — snapshot answering while the real backend swaps in — is covered by a dedicated test that asserts "exactly one response per id." That part is solid.

But there's a correctness gap in what the snapshot is allowed to answer from.

### Finding (HIGH, CONFIRMED): the snapshot fast path can serve a "0 files" answer for a project that simply hasn't been indexed yet, indistinguishable from a genuinely empty project

`src/daemon/router/session.ts`'s `buildSnapshotBackend()` (and the unused `SnapshotBackend.canOpen`, `src/daemon/router/snapshot-backend.ts:56`) gate the fast path on `fs.existsSync(sharedDbPath)` alone:

```ts
private buildSnapshotBackend(): SnapshotBackend | null {
  if (!fs.existsSync(this.opts.sharedDbPath)) return null;
  ...
}
```

But per the sibling fix that just merged in this same review batch (#956, `seedSessionDbFromShared` in `src/daemon/router/session-db.ts`), the shared DB file is created at **project registration**, before `indexAll()` ever populates it. `fs.existsSync` is true the instant the file exists — empty or not.

I reproduced this directly against `SnapshotBackend` (not the mocked test path): opened a DB via the same `initializeDatabase()` the PR's own tests use in `beforeEach` (0 rows in `files`), started `SnapshotBackend` against it, and called `get_project_map`:

```json
{"frameworks":[],"fileCount":0,"symbolCount":0,"languages":[],"_hints":[...]}
```

This is a normal, error-free, "truthy" response — exactly what the PR's own test at `src/daemon/router/__tests__/snapshot-backend.test.ts:76` asserts (`expect(call.result).toBeTruthy()`), so nothing in the test suite would catch it. A client that calls `get_project_map`/`search`/`get_outline` in the window between session bootstrap and `settleRealBackend()`'s swap — which is exactly the window this PR exists to serve traffic in — gets told the project has zero files and zero symbols, with no signal that this is "not indexed yet" rather than "actually empty." `settleRealBackend()` corrects the backend a moment later, but the wrong answer has already gone to the client and, for a real project, could plausibly be acted on (an agent concluding the repo is empty and changing course) before the correction arrives.

**Failure scenario:** first `serve` on a brand-new project (or any project whose shared DB was just recreated) — registration creates `sharedDbPath` synchronously; the daemon's `indexAll()` for a large repo can take seconds; a stdio session boots in that window, finds the file "exists," and answers the client's first `get_project_map`/`search` call from an empty snapshot instead of falling back to the (correct, if slower) bootstrap path this PR is trying to avoid taking unconditionally.

**Suggested fix:** reuse the same guard #956 just added — check that the shared DB actually has indexed rows (e.g. a `countIndexedFiles`-style check on `files`) before deciding the snapshot is "sitting on disk" and worth serving from, falling back to the normal `selectAndStartBackend()` path exactly the way `buildSnapshotBackend()` already does for "file doesn't exist." This is a small, surgical change to one gating function, not a redesign.

### Secondary, non-blocking

- `SnapshotBackend.canOpen()` (`src/daemon/router/snapshot-backend.ts:56`) is dead code — `buildSnapshotBackend()` reimplements the same `fs.existsSync` check inline instead of calling it. Worth collapsing to one place, especially since that's the exact function that needs the file-count check above.

### What I ran
- `npx vitest run src/daemon/router/__tests__/snapshot-backend.test.ts tests/daemon/session-snapshot-fast-path.test.ts` → 5 passed (doesn't cover the gap above — see finding)
- Direct repro against `SnapshotBackend` with an `initializeDatabase()`-only (0-file) DB, confirming the empty-but-successful response shown above

Not merging — please add the indexed-files guard to `buildSnapshotBackend()` and I'll take another look.

PR #1002, opened 2026-09-06
perf(graph): stop the per-file symbol N+1 in the daemon's graph build (TRA-651)

nikolai-vysotskyi:
Closes TRA-651.

## The reported number does not survive a settled measurement

TRA-651 filed *"serve-http holds 727 MB RSS **idle** while serving a single 1817-file project"*. Re-measured on v3.18.0 with the same pinned fixture (`fc47c10f44eb`, 1 817 files / 9 705 symbols), fresh `TRACE_MCP_DATA_DIR`, private port, exactly one registered project — sampled once a second with `heapUsed` alongside RSS:

| Daemon state | RSS | `heapUsed` |
|---|---|---|
| Fresh, nothing registered | 166 MB | — |
| First seconds after the index finishes | 700–720 MB | 200 MB |
| **Median over the next 60 s, untouched** | **399 MB** | **91 MB** |
| Median while a client drives it | 856 MB | ~300 MB |
| Peak while driven | 894 MB | 428 MB |
| 20 s after the client stops | 416 MB | 88 MB |

727 MB is a real reading, but it was taken inside the post-index window, not at idle. The settled idle figure is ~400 MB RSS over 91 MB of live heap, reached about twenty seconds after the last request. In one run RSS fell from 419 MB to 85 MB at t=80 s with `heapUsed` unchanged at 88 MB — the macOS page-retention effect TRA-278 documented, not a change in what is live. **There is no per-project leak to find here.**

What is real is the cost of *serving*, and it lands on the thread that answers `/health`.

## Where the serving cost comes from

`scripts/perf/daemon-query-alloc.mjs`, 200 calls per endpoint:

| Endpoint | ms/call | response | RSS growth / 200 calls |
|---|---|---|---|
| `/api/projects/graph` | **544** | **1 268 KB** | +99 MB |
| `/api/projects/smells` | **426** | 118 KB | **+293 MB** |
| `/api/projects/graph-stats` | 8.5 | 2.2 KB | +1 MB |
| `/api/projects/files` | 3.2 | 2.1 KB | +4 MB |
| `/health` | 0.14 | 0.2 KB | 0 |

`node --cpu-prof` over 12 whole-project `buildGraphData` calls put the time in two places, both work that did not need to happen:

1. **`getSymbolsByFileIds` was a per-file loop** despite the plural name — 424 ms self time. 1 507 statements to fetch 9 705 `SELECT *` symbol rows for a *file-level* graph that reads nothing but `id` and `file_id`; every signature and metadata blob in the project loaded and dropped.
2. **`getEdgesForNodesBatch` copied every row and re-parsed its SQL per chunk** — 442 ms self plus 136 ms in `prepare`. `{ ...row, pivot_node_id }` allocated a second object for every edge in the project on every build.

## The change

- `getSymbolsByFileIds` → one chunked `IN` query per 900 files.
- New `getSymbolFileRefsByFileIds` selects only `(id, file_id)`; the file-level graph path uses it.
- `getEdgesForNodesBatch` annotates `pivot_node_id` in place instead of spreading into a copy, and caches its chunk statement by placeholder count.

Nothing is cached at the response level and no result is memoized — same answer, fewer round trips and fewer allocations. Output is byte-identical before and after: 1 507 nodes / 5 070 edges / 922 738 bytes.

## Before → after

In-process, 12 consecutive whole-project builds:

| | before | after |
|---|---|---|
| Median build | 152 ms | **85 ms** (−44%) |
| `prepare` | 136 ms | 36 ms |
| GC | 264 ms | 67 ms |
| RSS after 12 builds | 415 MB | **307 MB** |

End-to-end on the daemon, 120 s of the app's query mix (`scripts/perf/daemon-serving-rss.mjs`, raw runs in `docs/perf/tra651-{before,after}.json`):

| | before | after |
|---|---|---|
| Queries served in 120 s | 1 492 | **2 150** (+44%) |
| Idle RSS after index | 399 MB | 404 MB |
| Idle `heapUsed` after index | 91 MB | 91 MB |
| Serving median RSS | 856 MB | 841 MB |
| Serving peak RSS | 894 MB | 865 MB |

**Resident memory while serving barely moved, and this PR is not claiming it did.** The daemon's RSS under load is set by GC pacing against a large young generation, not by how much each request allocates. The reason to take this change is the 44% latency and throughput win; the RSS number is flat and should not be quoted as a memory result.

## Guard

`tests/perf/graph-build-query-count.test.ts` asserts a whole-project graph build issues fewer statements than it has files. It counts statement *executions*, not `prepare` calls — the first version of this test counted prepares and passed on the broken code, because the N+1 reused a statement prepared in the repository constructor. Verified to fail on the pre-fix tree (128 queries for 120 files) and pass after. Machine-independent, no timing assertion to rot.

## Not fixed here

- `/api/projects/smells` at 426 ms and +293 MB per 200 calls is the larger remaining allocator on this path.
- Seven handlers in `src/cli.ts` open `new TopologyStore(TOPOLOGY_DB_PATH)` per request while `ProjectResourcePool` exists to hold one daemon-wide instance (TRA-938). 0.42 ms per open/close, so this is fd churn rather than latency — but it is the same descriptor-exhaustion shape TRA-938 fixed elsewhere.

Full write-up: `docs/perf/daemon-serving-memory.md`. Local: 10 252 tests pass, typecheck and `biome ci` clean.

### DIFF
--- a/docs/perf/daemon-serving-memory.md
+++ b/docs/perf/daemon-serving-memory.md
@@ -1,0 +1,130 @@
+---
+layout: default
+title: Daemon memory while serving one project
+permalink: /perf/daemon-serving-memory/
+description: Internal working document. What the trace-mcp daemon's resident memory does while a client drives it.
+noindex: true
+---
+
+# Daemon memory while serving one project
+
+TRA-651. Measured 2026-09-06 on darwin 25.5.0 / arm64, daemon v3.18.0 at `3e749a02`,
+throwaway `TRACE_MCP_DATA_DIR`, private port, **one** registered project: this repo at the
+pinned perf-fixture commit `fc47c10f44eb` — 1 817 files / 9 705 symbols. Reproduce with
+
+```
+pnpm run build
+node scripts/perf/daemon-serving-rss.mjs --drive-seconds 120 --idle-seconds 120
+node scripts/perf/daemon-query-alloc.mjs --calls 200
+```
+
+Raw runs: `docs/perf/tra651-before.json`, `docs/perf/tra651-after.json`.
+
+## The reported number does not survive a settled measurement
+
+TRA-651 filed "727 MB RSS **idle** while serving a single project". Sampling the same
+fixture on the same harness, at one-second resolution and with `heapUsed` alongside RSS:
+
+| Daemon state | RSS | `heapUsed` |
+|---|---|---|
+| Fresh, nothing registered | 166 MB | — |
+| First seconds after the index finishes | 700–720 MB | 200 MB |
+| **Median over the next 60 s, untouched** | **399 MB** | **91 MB** |
+| Median while a client drives it | 856 MB | ~300 MB |
+| Peak while driven | 894 MB | 428 MB |
+| 20 s after the client stops | 416 MB | 88 MB |
+
+727 MB is a real reading, but it is a reading taken inside the post-index window, not an
+idle steady state. The settled idle figure is ~400 MB of RSS over 91 MB of live heap, and
+it is reached about twenty seconds after the last request. Everything above that line is
+garbage V8 has not collected yet plus pages the process has not handed back — in one run
+RSS fell from 419 MB to 85 MB at t=80 s with `heapUsed` unchanged at 88 MB, which is the
+macOS page-retention effect TRA-278 documented and not a change in what is live.
+
+So there is no per-project leak to find here. What is real is the cost of *serving*: the
+daemon transiently doubles, and it does so on the same thread that answers `/health`.
+
+## Where the serving cost comes from
+
+`scripts/perf/daemon-query-alloc.mjs` drives one endpoint at a time, 200 calls each:
+
+| Endpoint | ms/call | response | RSS growth over 200 calls |
+|---|---|---|---|
+| `/api/projects/graph` | **544** | **1 268 KB** | +99 MB |
+| `/api/projects/smells` | **426** | 118 KB | **+293 MB** |
+| `/api/projects/graph-stats` | 8.5 | 2.2 KB | +1 MB |
+| `/api/projects/files` | 3.2 | 2.1 KB | +4 MB |
+| `/api/projects/stats` | 0.8 | 0.1 KB | ~0 |
+| `/health` | 0.14 | 0.2 KB | 0 |
+
+`/api/projects/graph` rebuilds the whole project graph from SQLite on every request — the
+app refetches it on any scope/granularity/filter change, and the result for identical
+parameters is byte-identical every time. 544 ms of that is synchronous `better-sqlite3`
+work on the process that also serves `/health`, which is the starvation shape from
+TRA-941.
+
+Profiling `buildGraphData` directly (12 consecutive whole-project builds,
+`node --cpu-prof`) put the time in two places, both of them work that did not need to
+happen:
+
+1. **`getSymbolsByFileIds` was a per-file loop** — 424 ms of self time. Despite the plural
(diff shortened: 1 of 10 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Eliminates the daemon graph-build N+1 (TRA-651): `getSymbolsByFileIds` went from one prepared statement per file (1,507 statements for a 1,817-file project) to one chunked query per 900 files; the chunked edge query now caches its prepared statement by placeholder count instead of re-parsing identical SQL on every 450-node chunk; a new `getSymbolFileRefsByFileIds` fetches only `(id, file_id)` for the file-level graph instead of full symbol rows (signatures/metadata) that were immediately discarded. Most of the diff (4,173 of 4,833 lines) is perf docs and before/after JSON baselines, not code.

### What I checked
- `edgesForNodesStmt`'s in-place row mutation (`row.pivot_node_id = ...` instead of `{...row, pivot_node_id}`) is safe: better-sqlite3's `.all()` returns freshly allocated row objects on every call, owned by nobody else — confirmed by reading the call site, no caching/pooling of row objects upstream.
- The per-chunk-size statement cache (`edgesForNodesCache: Map<number, Statement>`) is scoped to one `GraphRepository` instance (one per project DB), so its growth is bounded by the small number of distinct chunk sizes a single project's edge count actually produces over its lifetime (only the last, remainder chunk varies) — not a meaningful leak.
- Traced the one other caller of `Store.getSymbolsByFileIds` (`visualize.ts:623`) through to where its result is consumed: partitioning into `withEdges`/`withoutEdges` is done by `.filter()` against a `Set` membership check, not by relying on the input array's file order, so the new `ORDER BY file_id, byte_start` (numeric DB id order, not caller-array order) doesn't change which symbols get included when the `MAX_VIZ_NODES` cap kicks in — only their tie-break order within the excess slice, which was already an arbitrary Set-iteration order before this change.
- `getSymbolFileRefsByFileIds`'s trimmed `(id, file_id)` projection is safe for `buildFileGraph`: confirmed only `.id` (fed into `getNodeIdsBatch('symbol', ...)`) is used downstream, no other symbol field is read.
- `getEdgesForNodesBatch` (the method behind `edgesForNodesStmt`) is high blast-radius — used by `get_change_impact`, `get_call_graph`, `get_feature_context`, `get_task_context`, `get_tests_for` — so I ran the full breadth of tool tests that exercise it, plus the stress test that specifically drives it at 500 node IDs.

### Findings
None.

### What I ran
- `npx vitest run tests/perf/graph-build-query-count.test.ts` → 1 passed (new N+1 regression guard, asserts statement count stays far below file count)
- `npx vitest run tests/tools/visualize.test.ts tests/tools/visualize-aggregate.test.ts tests/tools/behavioural/visualize-graph.behavioural.test.ts tests/tools/behavioural/get-dependency-diagram.behavioural.test.ts tests/tools/behavioural/get-circular-imports.behavioural.test.ts tests/db/symbol-repository-fk-reinsert.test.ts tests/cli/visualize-open-in-browser.test.ts` → 53 passed
- `npx vitest run tests/tools/task-context.test.ts tests/tools/method-impact-repro.test.ts tests/tools/get_change_impact.test.ts tests/tools/call-graph.test.ts tests/tools/change-impact-output-budget.test.ts tests/tools/analysis/impact-churn-security.test.ts tests/tools/behavioural/get-{subproject,change,cross-workspace,federation}-impact*.test.ts tests/tools/behavioural/get-{call-graph,task-context,tests-for}.behavioural.test.ts tests/perf/stress.test.ts` → 147 passed

Merging.

PR #1003, opened 2026-09-06
feat(daemon): thin proxy entrypoint cuts connected-session RSS ~175MB→~73MB (TRA-970)

nikolai-vysotskyi:
## Summary

- New `dist/proxy.js` tsup entry (`src/proxy-entry.ts`): a minimal stdio↔daemon proxy process containing only the MCP SDK transport + `StdioSession`/`ProxyBackend` path — no Commander, no CLI command tree, no `LocalBackend`.
- `StdioSession.buildLocalBackend()` now dynamic-imports `local-backend.js` instead of a static import, and a new esbuild plugin (`localBackendExternal` in `tsup.config.ts`) marks that specifier `external` so it lands in its own `dist/local-backend.js` chunk instead of being inlined into every bundle that can reach it (proxy.js *and* cli.js).
- `subproject/resolve.ts`'s `better-sqlite3` usage is now a dynamic import too, for the same reason (it's reachable from `ProxyBackend.resolveProjectRoot()`).
- `hooks/trace-mcp-launcher.sh` now execs `proxy.js` directly instead of `cli.js` for a plain `serve` invocation (no args, or `serve [--preset X]`) when it can already see something listening on the daemon port — a cheap `/dev/tcp` heuristic; `proxy-entry.ts`'s own `StdioSession` still does the real `/health` check and falls back to local mode transparently if the shim guessed wrong. `cli.js`'s own behavior is untouched for every other invocation.
- Extracted the signal/stdin/parent-death shutdown wiring shared by `trace-mcp serve` and the new proxy entry into `runStdioSession()` (`src/daemon/router/session.ts`), so both processes exit identically instead of duplicating that logic.
- `LAUNCHER_VERSION` bumped 0.6.4 → 0.6.5 (all three shim headers) so existing installs pick up the new shim on next `trace-mcp init`/reinstall.

## Why dynamic import alone wasn't enough

With `splitting: false`, esbuild inlines a dynamically-imported *internal* module into the same bundle regardless — `dist/proxy.js` was still 11 MB after just switching `LocalBackend` to `await import(...)`. The fix was a small esbuild plugin that intercepts resolution of `./local-backend.js` and marks it `{ external: true }` before the catch-all `noExternal: [/.*/]` gets a say (mirroring what the existing `cjsViaCreateRequire` plugin does for native packages, for a different reason). That's what actually gets `dist/proxy.js` down to ~1 MB, backed by a sibling `dist/local-backend.js` chunk resolved at runtime the same way `extract-worker.js` already is.

## Test plan

- [x] `pnpm run build` / `pnpm run typecheck` / `pnpm run biome:ci` — all clean.
- [x] `pnpm run test` — full suite, 926 files / 10251 tests passed, 24 skipped, exit 0 (includes `tests/daemon/session*.test.ts`, `tests/daemon/proxy-backend.test.ts`, `tests/init/launcher-integration*.test.ts`, `src/subproject/__tests__/resolve.test.ts`).
- [x] Manually spawned `dist/proxy.js` against a live daemon on this machine: `initialize`/`tools/list` round-trip correctly; measured RSS via `ps -o rss=` at 5/10/20/30/45/60s — settled at **~73 MB at 60s** (target ≤70 MB; within run-to-run variance of the issue's own 69.8 MB reference figure for an isolated transport).
- [x] Manually exercised the launcher shim: plain `serve` with a daemon reachable → execs `proxy.js` (confirmed via its distinct log line); `search foo` → always `cli.js`; daemon-port heuristic pointed at a closed port → falls through safely to `cli.js`, which still connects to the real daemon on its own.
- [x] Confirmed `dist/local-backend.js` loads standalone and resolves correctly via a relative `import()` from `dist/proxy.js`'s own directory.
- Not independently re-verified: N=9 concurrent-session aggregate RSS and the daemon breakeven point (N*) — not remeasured on this box, but follow directly from the ~102 MB per-session reduction (175→~73 MB) already measured.

### DIFF
--- a/hooks/trace-mcp-launcher.cmd
+++ b/hooks/trace-mcp-launcher.cmd
@@ -1,5 +1,5 @@
 @echo off
-REM trace-mcp-launcher v0.6.6 (Windows)
+REM trace-mcp-launcher v0.6.7 (Windows)
 REM Tiny .cmd shim that invokes the PowerShell launcher. MCP clients spawn
 REM this .cmd because they rely on %PATHEXT% resolution which prefers .cmd.
 REM -WindowStyle Hidden keeps the cmd->powershell hop windowless: windowsHide
--- a/hooks/trace-mcp-launcher.ps1
+++ b/hooks/trace-mcp-launcher.ps1
@@ -1,4 +1,4 @@
-# trace-mcp-launcher v0.6.6 (Windows)
+# trace-mcp-launcher v0.6.7 (Windows)
 # Stable shim backend: resolves node + cli.js at runtime from launcher.env,
 # with a probe fallback for nvm-windows/nvs/Volta/system installs.
 # Managed by trace-mcp - do not edit by hand. Re-run `trace-mcp init` to refresh.
--- a/hooks/trace-mcp-launcher.sh
+++ b/hooks/trace-mcp-launcher.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# trace-mcp-launcher v0.6.6
+# trace-mcp-launcher v0.6.7
 # Stable shim: MCP clients invoke this path forever; it resolves node + cli.js
 # at runtime from a config file written by `trace-mcp init`, with a probe
 # fallback for when the config is stale (e.g. Node was reinstalled, or the
@@ -598,6 +598,55 @@ heal_config() {
   return 0
 }
 
+# --- TRA-970: daemon-aware fast path ---
+#
+# When a trace-mcp daemon is already listening, exec the thin proxy bundle
+# (dist/proxy.js, sibling of cli.js) instead of cli.js itself: it skips
+# loading Commander, PluginRegistry, better-sqlite3 and tree-sitter into this
+# process entirely, rather than loading all of it and only then discovering a
+# daemon exists — cli.js pays that cost just by being started, no matter which
+# subcommand runs.
+#
+# `/dev/tcp` only proves *something* is listening on the port, not that it is
+# a healthy trace-mcp daemon — deliberately cheap and approximate. proxy.js's
+# own StdioSession makes the real call (a proper /health request) and
+# transparently falls back to local mode if this guessed wrong (see
+# src/proxy-entry.ts and src/daemon/router/session.ts). A false positive here
+# costs a little startup latency, never correctness; a false negative just
+# skips the optimization for this one run.
+daemon_port_open() {
+  local port="${TRACE_MCP_DAEMON_PORT:-3741}"
+  ( : <"/dev/tcp/127.0.0.1/$port" ) 2>/dev/null
+}
+
+# True when argv is a plain `serve` invocation this shim understands well
+# enough to route to the thin proxy: no args, `serve`, or `serve --preset X`.
+# Anything else (serve-http, doctor, add, unrecognised flags, ...) always
+# takes the full cli.js path — Commander stays the source of truth for
+# everything this heuristic does not explicitly recognise.
+is_plain_serve() {
+  case $# in
+    0) return 0 ;;
+    1) [ "$1" = "serve" ] ;;
+    3) [ "$1" = "serve" ] && [ "$2" = "--preset" ] ;;
+    *) return 1 ;;
+  esac
+}
+
+# Echoes the script to exec: the thin proxy sibling of $1 when it exists, this
+# is a plain `serve` invocation, and a daemon is reachable — $1 (cli.js)
+# otherwise. Never errors; worst case it just echoes $1 back.
+resolve_exec_target() {
+  local cli="$1" proxy
+  shift
+  proxy="$(dirname "$cli")/proxy.js"
+  if [ -f "$proxy" ] && is_plain_serve "$@" && daemon_port_open; then
+    echo "$proxy"
+  else
+    echo "$cli"
+  fi
+}
+
 # --- 3. Fast path: config is good → exec directly ---
 #
 # A config recorded before the version gate existed carries no verified major.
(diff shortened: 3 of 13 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
Approved by Reviewer C (see the verdict comment above), but the merge now conflicts: #1001 (TRA-948) landed changes in `src/daemon/router/session.ts` and #985 (TRA-706) landed changes in `src/cli.ts` / `src/registry.ts` while this PR sat in the review-gate outage. Both are files this PR rewrites.

Needs a rebase onto current `master` by its author — the conflicts are in the proxy/session handoff, which is exactly the part a mechanical resolution would get wrong. Re-run the targeted tests after the rebase and it can merge; the review verdict stands for the change itself.

nikolai-vysotskyi:
## Verdict: APPROVE

Thin proxy entrypoint (TRA-970): the launcher shim now execs a new `dist/proxy.js` instead of `dist/cli.js` when it can cheaply tell a daemon is already listening and the invocation is a plain `serve`/`serve --preset X` — keeping Commander, PluginRegistry, better-sqlite3 and tree-sitter out of that process's module graph entirely. Falls back to `dist/cli.js` for anything else, and `StdioSession` degrades to `LocalBackend` (now dynamically imported) in place if the daemon turns out to be gone.

### What I checked
- Actually built the bundle (`npx tsup`) and inspected the output rather than trusting the tsup-config comments: `dist/proxy.js` is 1.06 MB vs `dist/cli.js` at 11.83 MB / `dist/local-backend.js` at 10.47 MB. Grepped both bundles for `import("./local-backend.js")` — confirmed it stays a real runtime `import()` in both (not inlined), which is the whole point of the new `localBackendExternal()` esbuild plugin in `tsup.config.ts`.
- Ran `dist/proxy.js` for real against this machine's live trace-mcp daemon (this sandbox shares the host's real daemon on :3741) via a stdio probe: it correctly used `ProxyBackend` when the daemon answered, and correctly fell back to `LocalBackend` (dynamic import kicking in, full stack loading only then) in a run where the daemon didn't respond in time — both are the two documented outcomes, both worked.
- All 4 call sites of the now-async `buildLocalBackend()` in `session.ts` are properly `await`ed (checked each one, not just grepped for the word `await`), and the 3 call sites of the now-async `resolveDeepestKnownRoot`/`isKnownSubproject`/`findSubprojectRootForPath` chain in `cli.ts`/`proxy-backend.ts`/`resolve.ts` likewise.
- **Manually verified the new shell logic end-to-end**, since nothing in the diff or the existing test suite exercises it: extracted `daemon_port_open`/`is_plain_serve`/`resolve_exec_target` from `hooks/trace-mcp-launcher.sh` and ran them against a real listening socket for every branch — bare invocation, `serve`, `serve --preset dev`, `serve --foo` (unrecognized 2-arg form), `serve --preset` alone (missing value), `serve-http`, `doctor`, and "no `proxy.js` sibling." Every case routed exactly where the code comments say it should: plain `serve`/`serve --preset X` → proxy.js when a daemon is listening, everything else (including any 2-arg form the heuristic doesn't explicitly recognize) → cli.js. The `/dev/tcp` check only proves *something* is listening, never that it's a healthy trace-mcp daemon — correct, since `proxy-entry.ts`'s own `StdioSession` makes the real `/health` call and falls back safely either way, exactly as documented.
- The Windows launcher (`.cmd`/`.ps1`) gets only a version bump in this PR, consistent with the existing precedent (e.g. the mirror hook) of shipping Unix-first — not a gap introduced silently, the diff doesn't claim Windows support.
- `getEdgesForNodesBatch`-adjacent worry doesn't apply here; checked the actual blast radius of this PR instead: `getEdgesForNodesBatch` wasn't touched. High-blast-radius here is `StdioSession`/`ProxyBackend`, so I ran the daemon/cli/init test suites broadly, not just the new/changed files.

### Findings (non-blocking)
- **Test-coverage gap:** `resolve_exec_target`, `is_plain_serve`, and `daemon_port_open` in `hooks/trace-mcp-launcher.sh` have no automated coverage — `tests/init/launcher-integration.test.ts` already spawns the real script as a subprocess for other scenarios (fake `cli.js`, fake `HOME`), so adding a fake `proxy.js` sibling + a real listening socket fixture would have been straightforward and would exercise the actual new logic instead of only confirming (as the existing 48 tests do) that behavior is unchanged when `proxy.js` doesn't exist. I verified the logic manually and it is correct in every case I tried, but this is exactly the kind of shell change (executed by every MCP client on every session start) that benefits from a regression guard living in the repo rather than in a reviewer's scratch script.

### What I ran
- `npx tsup` → real build, inspected `dist/proxy.js` / `dist/cli.js` / `dist/local-backend.js` sizes and import wiring
- `npx vitest run src/subproject/__tests__/resolve.test.ts tests/daemon/session-cold-start-handshake.test.ts tests/daemon/local-backend-reentry.test.ts tests/daemon/local-backend-dangerous-root.test.ts tests/daemon/local-backend-readonly.test.ts` → 25 passed
- `npx vitest run tests/daemon tests/cli tests/init` → 690 passed, 7 skipped
- Manual shell-function extraction + real-socket test of `resolve_exec_target`/`is_plain_serve`/`daemon_port_open` (see above) — all branches correct

Merging.

nikolai-vysotskyi:
Review above stands (APPROVE). Merge itself is blocked: GitHub reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING` against current `master` — this branch predates several PRs just merged in this same review pass (#985, #993, #999, #1002, all of which touch `src/cli.ts` / `src/daemon/router/session.ts` / `src/registry.ts`), so the conflict is fallout from merge order, not a problem with this diff. Needs a rebase onto current `master` before it can go in — not something I'll do unilaterally on someone else's branch. Leaving unmerged for the author to rebase.

PR #1004, opened 2026-09-06
fix(launcher): name control-plane tmp files so the orphan sweeper collects them (TRA-982)

nikolai-vysotskyi:
## What

Two writers put files into the trace-mcp state home (and its `bin/` directory) with an atomic write, but named their tmp file in a shape the orphan sweeper cannot match — so a process killed between the write and the rename leaked a file that nothing would ever collect.

`sweepOrphanTmpFiles` matches `^\.?.+\.tmp\.\d+\.[0-9a-f]{12}$` and `bin` is in `SWEPT_STATE_DIRS` precisely so leftovers next to the launcher shim get cleaned up. The trailing hex is deliberate — it keeps the sweeper from eating a user file that merely has `.tmp.` in its name.

| Writer | Emitted | Swept? |
|---|---|---|
| `atomicWriteString` (`src/utils/atomic-write.ts`) | `.tmp.<pid>.<12 hex>` | yes |
| `acquireLock` (`src/utils/pid-lock.ts`) | `.tmp.<pid>.<12 hex>` | yes |
| `writeLegacyCompat` (`src/init/launcher.ts`) | `.tmp.<pid>` | **no** |
| `writeIfChanged` (`packages/app/src/main/daemon-install.ts`) | `.tmp.<pid>.<epoch ms>` (13 decimal digits) | **no** |

Both now use `randomBytes(6).toString('hex')`.

## How it was found

TRA-982 autopilot run. `claude mcp list` shows trace-mcp connected and `launcher.log` has zero `ERROR` lines in the last 24h, so the run moved on to the untested failure scenarios the issue lists. Item 5 (state hygiene — "остаются осиротевшие `.tmp.*` от прерванных атомарных записей") led to auditing every tmp-name producer against the sweeper's pattern.

## Tests

Both tests capture the tmp path the writer actually hands to `renameSync`/`symlinkSync`, so they pin the producer rather than restating the regex:

- `tests/init/launcher-legacy-compat.test.ts` — writes a leftover with the captured name into a temp dir, ages it, and asserts a real `sweepOrphanTmpFiles` pass collects it.
- `packages/app/src/main/__tests__/daemon-install.test.ts` — asserts every tmp name a full `ensureDaemonInstalled` produces matches the sweeper's pattern. (The regex is duplicated there rather than imported: `packages/app` is its own package with its own test setup and does not import from the server's `src`.)

Both fail on `main` (`trace-mcp.tmp.48403`, `node-runtime.tmp.48599.1788660433641`) and pass with the fix.

Full suite: `927 files / 10275 tests passed`, `pnpm typecheck`, `pnpm biome:ci`, `pnpm build` all clean. The `packages/app` suite has 44 pre-existing file-level failures (react/jsx-runtime resolution) — identical counts with and without this change; it adds one passing test (233 → 234).

## Not in this PR

The run also reproduced the TRA-965 dangling-`node-runtime`-shim outage against the copy installed on this machine (launcher v0.6.3: `exit 126`, all ~170 tools lost, no `ERROR` line). The fix is already on `main` as launcher v0.6.5 and recovers correctly — it is simply unreleased (repo 3.19.0 vs 3.18.0 installed everywhere). Release timing, not a code defect.

Closes TRA-982.

### DIFF
--- a/packages/app/src/main/__tests__/daemon-install.test.ts
+++ b/packages/app/src/main/__tests__/daemon-install.test.ts
@@ -5,7 +5,7 @@
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import net from 'node:net';
 import {
   compareVersions,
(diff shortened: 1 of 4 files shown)


nikolai-vysotskyi:

nikolai-vysotskyi:
## Verdict: APPROVE

Fixes two atomic-write tmp-file naming spots (`packages/app/src/main/daemon-install.ts`'s `writeIfChanged`, `src/init/launcher.ts`'s `writeLegacyCompat`) that used a shape (`.tmp.<pid>` / `.tmp.<pid>.<epoch-ms>`) the server's own orphan sweeper (`ORPHAN_TMP_PATTERN` in `src/utils/atomic-write.ts`, `.tmp.<pid>.<12 hex>`) never matched — so a process killed mid-write left a permanent, uncollected file next to the launcher shim or control-plane files.

### What I checked
- Confirmed the regex both writers now target matches the real sweeper pattern exactly: `ORPHAN_TMP_PATTERN = /^\.?.+\.tmp\.\d+\.[0-9a-f]{12}$/` in `src/utils/atomic-write.ts:142`; `randomBytes(6).toString('hex')` produces exactly 12 hex chars.
- Confirmed `'bin'` is in `SWEPT_STATE_DIRS` (`src/utils/atomic-write.ts:196`), so the launcher shim's directory is actually swept, and that `packages/app` deliberately duplicates the regex in its own test rather than importing across the app/server package boundary — reasonable given they're separate packages.
- `packages/app`'s new test asserts the emitted tmp name against a locally duplicated copy of the pattern; `src/init/launcher.ts`'s new test goes further and round-trips the name through the real `sweepOrphanTmpFiles()` to prove it's actually collected, not just pattern-shaped — the stronger check of the two, and it's the one testing the real sweeper.

### Findings
None.

### What I ran
- `npx vitest run tests/init/launcher-legacy-compat.test.ts` → 15 passed
- `cd packages/app && npx vitest run src/main/__tests__/daemon-install.test.ts` → 30 passed

Merging.

PR #1014, opened 2026-09-06
ci: cache pnpm activation via corepack instead of re-downloading per job

nikolai-vysotskyi:
## Summary

- `pnpm/action-setup`'s self-installer downloads pnpm from the npm registry on every job that needs it (measured 84s-6m28s depending on registry health) — root cause of the TRA-785 outage where a slow registry cancelled `audit` mid-setup and blocked master for ~4h.
- Replaces it with a local composite action (`.github/actions/setup-pnpm`) that activates the pinned pnpm version via Node 22's built-in corepack, caching corepack's download directory keyed on the pnpm version string (not the lockfile, so dependency bumps don't bust it).
- `impact-report`, `quality-gates-report`, `security-scan-report` never actually invoked `pnpm` (they only run `node dist/cli.js` against the prebuilt artifact) — pnpm setup is removed there entirely instead of replaced.
- `cross-platform-test` (macOS/Windows, rarely triggered) is deliberately left on `pnpm/action-setup` — not the everyday cost this targets, and the riskiest runners to change blind.
- No change to any job's commands, to the `audit` gate semantics from #853, or to the `timeout-minutes` ceilings from #850.

Closes TRA-790.

## Local verification

Confirmed the mechanism directly (Node v22.22.3, corepack 0.34.6):
```
$ export COREPACK_HOME=/tmp/corepack-test-home; rm -rf "$COREPACK_HOME"
$ corepack enable
$ time corepack prepare --activate      # cold: 1.6s (network fetch)
$ time corepack prepare --activate      # warm: 0.046s (cache hit, no fetch)
$ pnpm --version
10.33.0
```

## Test plan

- [x] `yaml.safe_load` on both changed/added files
- [x] Local corepack cold→warm cache round-trip (above)
- [ ] Push to this branch shows the pnpm setup step as a cache hit on jobs after the first — CI timings below once this run completes
- [ ] Second-model review (red-zone: `.github/workflows/**`)

### DIFF
--- a/.github/actions/setup-pnpm/action.yml
+++ b/.github/actions/setup-pnpm/action.yml
@@ -1,0 +1,37 @@
+name: Setup pnpm (corepack, cached)
+description: >-
+  Activates the pnpm version pinned in package.json's "packageManager" field
+  via Node's built-in corepack, and caches corepack's download directory so
+  the pnpm tarball is fetched from the npm registry at most once per pnpm
+  version bump instead of once per job. Requires actions/setup-node (Node
+  >=22) to have already run.
+runs:
+  using: composite
+  steps:
+    # Job-wide (not just this step) via $GITHUB_ENV, so every later `pnpm`
+    # invocation in the calling job reads from the same cached directory
+    # instead of corepack's default per-user location.
+    - name: Point corepack's download cache at a cacheable path
+      shell: bash
+      run: echo "COREPACK_HOME=${{ runner.temp }}/corepack-home" >> "$GITHUB_ENV"
+
+    - name: Enable corepack
+      shell: bash
+      run: corepack enable
+
+    - name: Resolve pinned pnpm version
+      id: pnpm-version
+      shell: bash
+      run: echo "value=$(node -p "require('./package.json').packageManager")" >> "$GITHUB_OUTPUT"
+
+    # Keyed on the pnpm version string, not on the lockfile or package.json
+    # hash — a dependency bump shouldn't bust this, only a pnpm upgrade should.
+    - name: Cache pnpm download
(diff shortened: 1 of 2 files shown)


nikolai-vysotskyi:
## Before/after: pnpm setup timing

Confirmed across two consecutive real runs on this branch (both `biome`, ubuntu-latest):

**Run 1** (https://github.com/nikolai-vysotskyi/trace-mcp/actions/runs/34023122851/job/101459231990) — first-ever use of the new cache key:
```
Cache not found for input keys: corepack-Linux-pnpm@10.33.0
Preparing pnpm@10.33.0 for immediate activation...
[...] duration_ms=1244   # registry was healthy; historically 84s-6m28s per TRA-785
```

**Run 2** (https://github.com/nikolai-vysotskyi/trace-mcp/actions/runs/34023637763/job/101460646674) — same pnpm version, next run:
```
Cache hit for: corepack-Linux-pnpm@10.33.0
Preparing pnpm@10.33.0 for immediate activation...
[...] duration_ms=73   # no registry round-trip at all
```

Same pattern held for the pnpm-store cache setup-node still manages for `biome`/`vscode-extension` (`node-cache-Linux-x64-pnpm-<hash>`, hit on run 2). One real bug caught along the way: `build` and `app-typecheck` install conditionally on their own node_modules cache, so a cold pnpm-store cache there (new physical path under corepack, since it doesn't set `PNPM_HOME` the way `pnpm/action-setup` did) skipped install entirely and left the store directory missing — `actions/cache`'s post-job save hard-errored on a nonexistent path. Fixed by dropping `cache: 'pnpm'` from those two jobs, since their own node_modules cache already covers the case it existed for; `biome`/`vscode-extension` keep it since their installs are unconditional.

All CI green on the current HEAD (08e583b2): https://github.com/nikolai-vysotskyi/trace-mcp/pull/1014/checks

PR #990, opened 2026-09-06
docs(seo): route `mcp tracing` impressions to /telemetry.html (TRA-974)

nikolai-vysotskyi:
## Problem

GSC, 2026-08-07 → 2026-09-05: `mcp tracing` gives **54 impressions, avg position 13.7, CTR 0** — 28% of all attributed impressions. The page breakdown shows they go to the wrong page:

| Page | Impressions |
|---|---|
| homepage | ~51 |
| `/configuration.html` | 2 (position 70.5) |
| `/telemetry.html` | 1 |

`/telemetry.html` targets that exact phrase since 2026-05-13 and is `Submitted and indexed` (URL Inspection 2026-09-06, last crawl 2026-08-28). Internal cannibalization: the homepage outranks it while not discussing tracing at all.

**Cause:** `/telemetry.html` had *zero* in-body inbound links — only the flat footer nav. The one homepage link that looked like an exception (`href="telemetry"` in the savings note) is inside `{% if site.data.savings %}`, and `docs/_data/savings.yml` does not exist, so it never rendered — and the href was relative-broken regardless.

## Change

- Contextual in-body links with anchor text **"MCP tracing"** from the three indexed pages that can carry one: `index.html` (tools section), `configuration.md`, `analytics.md`.
- `href="telemetry"` → `/telemetry.html`.
- `ops/index-coverage.md`: the breakdown, the fix, and the GSC verification method — impressions should move off the homepage onto `/telemetry.html`, position should beat 13.7.
- Correction recorded in `ops/rename-to-trace.md`: `mcp tracing` is **our** category (we emit OTel spans per MCP tool call), not an out-of-category collision as `docs/ROADMAP.md` item 19 states. `traceix mcp` (61 impressions, position 6.0, CTR 0) is the real foreign-brand collision — deliberately left alone.

Docs-only; `llms-full.txt` regenerated with `pnpm docs:sitemap` (sitemap dates unchanged).

## Test evidence

- `npx vitest run tests/docs` — 19 files, 248 passed, 4 skipped.
- `npx vitest run` (full) — 924 files, 10249 passed, 26 skipped.

## Follow-up (not in this PR)

Request Indexing for `/telemetry.html` in GSC so the recrawl isn't a month out — one click, batched with TRA-633.

Refs TRA-974

PR #1025, opened 2026-09-06
docs(seo): link tools-reference to MCP tracing; record the TRA-1025 re-derivation

nikolai-vysotskyi:
Closes TRA-1025 — which re-derives TRA-974 (merged in #990 a few hours earlier) from the same GSC window. Two of its three asks were already shipped or have no target; this PR lands the one that did.

## What changed

**`docs/tools-reference.md`** — one contextual in-body link to `telemetry.md` with the anchor "MCP tracing", in the Session Analytics section. TRA-1025's page breakdown surfaced a fourth page carrying the `mcp tracing` query that #990's table missed: `/tools-reference.html`, 2 impressions at position 92. It was the only one of the four indexed pages taking the query with no in-body link to the profile page (#990 covered the homepage, `configuration.html` and `analytics.html`).

**`ops/index-coverage.md`** — appended to the TRA-974 section so the next run doesn't file this a third time:

- The fourth page, and that the `/#capabilities`, `/#install`, `/#problem`, `/#product` rows in TRA-1025's table are the same homepage impressions counted per in-page anchor, not extra pages.
- **"Remove the tracing wording from the homepage" has no target.** The string `tracing` appears exactly once in `docs/index.html` (line 2705) and it is the outbound anchor to `/telemetry.html` that #990 added. The homepage wins this query on domain/entity match (`trace-mcp` ≈ `mcp tracing`) plus host crowding, not on copy.
- **CTR 0 is not evidence here**, and both issues leaned on it. 51 impressions at position 8.3 has an expected yield of ~1 click; zero is inside the noise floor. The load-bearing fact is the page split (1 impression to the profile page against the homepage's 51), not the CTR — same trap as the `/vs/` inference the file already corrects two sections down.

`docs/llms-full.txt` and `docs/sitemap.xml` are regenerated by `pnpm docs:sitemap`, not hand-edited.

## Test evidence

`pnpm exec vitest run tests/docs` — 20 files, 266 passed / 4 skipped, including `internal-links`, `readme-claims` and the sitemap `lastmod` guards.

No source, tool-signature or schema changes; token cost unaffected.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -156,7 +156,9 @@ Decisions auto-enrich code intelligence: `get_change_impact` shows `linked_decis
 
 ## Session Analytics
 
-See [Analytics](analytics.md) for full documentation.
+See [Analytics](analytics.md) for full documentation. Those tools read session
+logs after the fact; for a live span per tool call exported to your own Jaeger
+or Langfuse, see [MCP tracing](telemetry.md).
 
 | Tool | What it does |
 |---|---|
--- a/docs/sitemap.xml
+++ b/docs/sitemap.xml
@@ -8,7 +8,7 @@
   </url>
   <url>
     <loc>https://trace-mcp.com/tools-reference.html</loc>
-    <lastmod>2026-09-04</lastmod>
+    <lastmod>2026-09-06</lastmod>
     <changefreq>monthly</changefreq>
     <priority>0.8</priority>
   </url>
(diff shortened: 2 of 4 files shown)


nikolai-vysotskyi:
Merging without a second-model review under the documented exception in the workspace rules — «правка текста в markdown-файле». The diff is two markdown files plus `docs/llms-full.txt` and `docs/sitemap.xml` regenerated by `pnpm docs:sitemap`; no source, no CI workflow, no on-disk schema, no MCP tool contract.

PR #1028, opened 2026-09-06
fix(indexer): resolve C# using-directive imports (TRA-1027)

nikolai-vysotskyi:
## Summary

`import-capable-languages.ts` named C# explicitly: the plugin extracts `using` directive edges (`extractImportEdges` in `csharp/helpers.ts`) but nothing consumed them, so a C# repo indexed with zero import edges regardless of what the matrix claimed. Same class of gap TRA-483 (Java) and TRA-900 (Ruby) closed before it.

- Adds `resolveCSharpImportEdges` (`src/indexer/edge-resolvers/csharp-imports.ts`), wired into the pipeline after Ruby.
- Unlike Java/Go, C# namespaces don't have to mirror the directory layout, so resolution matches the specifier against declared `namespace` symbols (already extracted by `CSharpLanguagePlugin.extractNamespace`) rather than path suffixes. Covers plain `using X.Y;`, `using static X.Y.Type;`, and aliased `using A = X.Y.Type;` (trims to the namespace).
- Moves `csharp` from the "extracts but nothing consumes" list to `IMPORT_EDGE_LANGUAGES`; regenerated `docs/language-matrix.md` (Imports 18 → 19) and `docs/llms-full.txt`.

## Verification

- New E2E test (`tests/integration/csharp-import-resolution-e2e.test.ts`), modeled on the Java one — plain namespace import, `using static`, alias, and BCL imports correctly left unresolved.
- Full suite: `pnpm run build` clean, `pnpm run lint` clean, `pnpm run test` — 10,474 passed, 0 failed.
- Checked against a real public repo (Newtonsoft.Json, 945 `.cs` files, not a fixture): 305 files gained at least one resolved import edge, 26,347 edges resolved total, 1,713 external (BCL) specifiers correctly left unresolved rather than inventing nodes.

## Test plan

- [x] `pnpm run build`
- [x] `pnpm run lint`
- [x] `pnpm run test` (full suite, 0 failures)
- [x] Indexed a real C# repo (Newtonsoft.Json) and manually inspected resolved edges

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,7 +1,7 @@
 ---
 title: "Contributing to trace-mcp — local setup, build, and test"
 description: "How to set up trace-mcp for local development: install, build, run the test suite, and the conventions to follow before opening a PR."
-updated: 2026-09-04
+updated: 2026-09-06
 ---
 
 # Development
--- a/docs/language-matrix.md
+++ b/docs/language-matrix.md
@@ -92,6 +92,17 @@ without one the extracted import never becomes an edge. This column counts the
 second thing, so it is much shorter than the language list — see
 `src/indexer/edge-resolvers/import-capable-languages.ts`.
 
+**"Yes" is not "every import."** C# is the narrowest case in this column today:
+its resolver only trusts `using` forms that name a specific type (`using
+static X.Y.Type`, an aliased `using A = X.Y.Type`) and deliberately leaves a
+plain `using Namespace;` — the overwhelming majority of real C# imports —
+unresolved, because a namespace can span most of a codebase and a file-level
+edge to all of it would be closer to noise than signal. On a large real repo
+(Newtonsoft.Json, ~5,000 `using` directives) this column's "yes" for C#
+produced edges for 12 of them — roughly 0.2%. See
+`src/indexer/edge-resolvers/csharp-imports.ts` for the reasoning and
+`src/indexer/plugins/language/csharp/helpers.ts` for extraction.
+
 ## Matrix
 
 | Language | Extensions | Parser | Default | Imports | Calls | Types | Tests |
--- a/docs/language-matrix.md
+++ b/docs/language-matrix.md
@@ -48,7 +48,7 @@ is the two-pass pipeline in [architecture](architecture.md#indexing-pipeline).
 
 - **indexed with the default config:** 78 (the rest need an `include` entry — see below)
 - **tree-sitter parser:** 29 · **regex parser:** 46 · **custom parser:** 6
-- **import edges:** 18
+- **import edges:** 19
 - **call edges (`get_call_graph`, call-aware `find_usages`):** 3
 - **type / inheritance edges:** 2
 - **covered by a plugin test:** 68
@@ -110,7 +110,7 @@ second thing, so it is much shorter than the language list — see
 | cobol | .cob .cbl .cpy .cobol | regex (multi-pass) | yes | — | — | — | yes |
 | common-lisp | .lisp .lsp .cl .asd | regex (multi-pass) | yes | — | — | — | yes |
 | cpp | .cpp .cxx .cc .hpp | tree-sitter | yes | yes | — | — | yes |
-| csharp | .cs | tree-sitter | yes | — | — | — | yes |
+| csharp | .cs | tree-sitter | yes | yes | — | — | yes |
 | css | .css .scss .sass .less | tree-sitter | yes | yes | — | — | yes |
 | cuda | .cu .cuh | regex | yes | — | — | — | — |
 | d | .d .di | regex (multi-pass) | yes | — | — | — | yes |
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -4784,7 +4784,7 @@ is the two-pass pipeline in [architecture](architecture.md#indexing-pipeline).
 
 - **indexed with the default config:** 78 (the rest need an `include` entry — see below)
 - **tree-sitter parser:** 29 · **regex parser:** 46 · **custom parser:** 6
-- **import edges:** 18
+- **import edges:** 19
 - **call edges (`get_call_graph`, call-aware `find_usages`):** 3
 - **type / inheritance edges:** 2
 - **covered by a plugin test:** 68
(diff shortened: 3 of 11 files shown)


nikolai-vysotskyi:
Thanks for the review — pushed a fix in b85f1abf that follows your suggested replacement closely:

- **Finding 1** (unindexed sub-namespace treated as parent): fixed. `resolve()` now only falls back by trimming into the new `byType` map, never back into `byNamespace` — a miss on a real namespace (e.g. `using Acme.Store.Missing;`) is external, full stop.
- **Finding 2** (`using static`/aliased type fanning out to the whole namespace): fixed. Type symbols (`class`/`interface`/`enum`/`type`) are now indexed into their own `byType` map (`FQN → file ids`), and `resolve()` tries an exact type match before falling back to the trim.
- **Finding 3** (duplicate file ids inflating `created`): fixed via a shared `addTo()` helper that dedupes before pushing.
- **Finding 4** (global/nested types under `using static`): covered by the same change — direct `byType` lookup handles a dotless global type, and the trim-into-`byType` fallback handles nested types.

Extended the E2E test with a namespace spanning two files (so a type-level `using` can be checked against fanning out) and an unindexed sub-namespace (so the trim fallback can be checked against the wrong-parent match). `pnpm run build`, `pnpm run lint`, full `pnpm run test` (10,475 passed) all clean.

Re-verified against the same real repo (Newtonsoft.Json): edge count dropped slightly (26,461 → 26,291) as the false-positive shapes above were removed, while the dominant edge shape (plain `using Namespace;` fanning out to every file that legitimately shares that namespace) is unaffected — that fan-out is correct C# semantics, not a bug.

nikolai-vysotskyi:
Fixed in 3c19d093, addressing Code Reviewer's second-round findings:

- **High (plain namespace fan-out):** a plain `using Namespace;` now only resolves when exactly one file declares that namespace — the same precision every other import in this codebase gets. More than one declaring file is left unresolved rather than guessed. Re-verified against the same real repo (Newtonsoft.Json): resolved edges dropped from 26,461 to 12 (avg fan-out 1.0, max 1) — the vast majority of that repo's `using` directives name namespaces shared by many files, which is exactly the ambiguous case this now correctly declines rather than guesses at. The fully precise fix (resolve only the types actually referenced in a file's body) is real usage-tracking and is out of scope here — noted as a follow-up for a future run of this topic.
- **High (stale edge on namespace rename):** switched the incremental-reindex driver from `state.pendingImports` (only has entries for files with their own `using` directives) to `state.changedFileIds` (every re-extracted file), and added a pass that re-validates a changed C# file's incoming `imports` edges against the fresh namespace/type maps, deleting any that no longer resolve. Regression test reproduces the exact scenario (rename `Store.cs`'s namespace, reindex only that file, assert the importer's stale edge is gone).
- **Medium (namespace-scoped `using` not extracted):** the plugin now collects `using_directive` nodes recursively instead of only at the compilation-unit root, so `namespace X { using Y; ... }` is no longer silently dropped.

`pnpm run build`/`lint` clean, full `pnpm run test` — 10,478 passed, 0 failed.

nikolai-vysotskyi:
Agreed on both findings — pushed `75a2a042`.

Rather than patch the cardinality-transition bug, I took your suggested minimal fix: removed the `byNamespace` map entirely. A plain `using Namespace;` is now always left unresolved, regardless of how many files declare it. Only forms naming a specific type — `using static X.Y.Type`, aliased `using A = X.Y.Type` — still resolve, via the unchanged `byType` map. This removes the whole class of bug (nothing to transition between anymore) and the semantic imprecision you flagged in finding 2, at the cost of a real capability narrowing: most `using` directives in real C# code are plain namespace imports, so this resolver will produce far fewer edges than it did after the first fix. That's an honest tradeoff given the alternative was an unsound one.

Re-verified against Newtonsoft.Json: the 12 type-precise edges are unchanged from the previous round (they were never namespace-based, so removing `byNamespace` didn't touch them). `pnpm run build`/`lint` clean, full `pnpm run test` — 10,477 passed, 0 failed.

nikolai-vysotskyi:
Fixed both in `5fde36eb`.

1. **High (move doesn't converge without a full reindex):** the revalidation pass now relinks, not just prunes. A stale incoming edge already carries its own `source_node_id` and stored `metadata.from` — enough to re-resolve against the fresh type map and insert an edge to wherever the type lives now, without re-extracting the importer. Added a regression test: type moves from `One.cs` to `Two.cs` (both reindexed in the same incremental batch, as a file watcher would), importer untouched, `App.cs -> One.cs` is gone and `App.cs -> Two.cs` exists. The one case that still doesn't converge without a full reindex — a specifier that was external when the importer was last extracted, if the type it names shows up later in some file — is now called out explicitly in the resolver's file header rather than left implicit.
2. **Medium (matrix overstates the capability):** added a caveat paragraph to `scripts/language-matrix.ts`'s generator naming C# specifically — "yes" in the Imports column doesn't mean "every import," and the real-world number (12 of ~5,000 `using` directives in Newtonsoft.Json, ~0.2%) is stated directly rather than left for a reader to discover the hard way.

`pnpm run build`/`lint` clean, full `pnpm run test` — 10,478 passed, 0 failed.

nikolai-vysotskyi:
Fixed both in \`594beb71\`.

1. **Medium (multi-specifier collapse loses metadata):** reworked around \`metadata.froms: string[]\` instead of a single \`from\`. Both passes (revalidation + source-driven) now build an in-memory (source, target) → specifier-set map first, and write once per edge at the end — nothing gets clobbered regardless of pass order. Regression test: two \`using static\` imports into the same file, moving only one, the other stays.
2. **High (delete+create rename doesn't converge):** confirmed and pinned rather than silently left implicit. A real watcher rename goes through \`deleteFiles\` then a separate \`indexFiles\` call, and \`deleteFiles\` cascades away the edge this resolver's relink logic depends on before the create is seen — there's genuinely nothing left to recover from without raw import facts surviving independently of the resolved edge, which is real persistence work spanning every per-language import resolver (Java/Go/Ruby/Rust/C likely have some version of this too — nobody's tested it there either). Filed TRA-1038 for that, assigned Lead Engineer, backlog/not-urgent — the current behavior is honest (missing edge, not a wrong one) and self-heals on a full reindex, both proved by a new test.

\`pnpm run build\`/\`lint\` clean, full \`pnpm run test\` — 10,480 passed, 0 failed.

nikolai-vysotskyi:
Fixed in \`3cb155f0\` — the reported bug was exactly the early return, nothing more: removed \`if (byType.size === 0) return;\`. The rest of the function already handles an empty \`byType\` correctly (\`resolve()\` returns \`[]\` for everything, which is the right answer for both passes). Added a regression test using C# top-level statements to reach the specific shape needed (an importer that itself declares no type, so the target losing its only type makes \`byType\` empty across the whole index).

Also tightened TRA-1038's description to say "forced full reindex" throughout, per your note.

\`pnpm run build\`/\`lint\` clean, full \`pnpm run test\` — 10,480 passed (one unrelated flaky hook-script test, \`tests/hooks/precompact-hook.test.ts\`, passed cleanly on its own re-run). Re-verified against Newtonsoft.Json: still 12 resolved edges, unchanged.

PR #1050, opened 2026-09-06
fix(app): disambiguate identically-named projects in Recent/Quick Open (TRA-1058)

nikolai-vysotskyi:
## Summary
- Fixes the #1-priority finding from Lead Engineer's QA pass (TRA-1058, comment https://multica.ai): recent-project rows and the Quick Open recent group render only the basename, so two roots that share a folder name (e.g. multiple agent `.../workdir` checkouts) are indistinguishable — reported as 7 of 43 rows non-unique, 5 in a row named `workdir`.
- Adds `disambiguateProjectLabels()` (`recent-projects.ts`): grows each colliding label one parent path segment at a time until the set is unique, the way Finder and IDE tabs disambiguate duplicate names.
- Wires it into the sidebar's `RecentProjects` list (reusing the existing `.ws-sb-path .name/.dir` styling already used for file rows) and into the Quick Open recent-project group.

## Test plan
- [x] `pnpm run typecheck` — clean
- [x] `pnpm run test` — 724/724 passing (new `recent-projects.test.ts` covers unique names, one-level collisions, multi-level collisions, and the "still ambiguous, no crash" exhausted case)

### DIFF
--- a/packages/app/src/renderer/App.tsx
+++ b/packages/app/src/renderer/App.tsx
@@ -129,6 +129,7 @@ const BASE = 'http://127.0.0.1:3741';
 import { daemonFetch } from './daemon-fetch';
 import {
   addRecentProject,
+  disambiguateProjectLabels,
   getRecentProjects,
   removeRecentProject,
 } from './recent-projects.js';
@@ -178,6 +179,8 @@ function RecentProjects() {
     setRecent(getRecentProjects());
   };
 
+  const recentLabels = disambiguateProjectLabels(recent);
+
   return (
     <>
       {ctx && (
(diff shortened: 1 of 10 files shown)


nikolai-vysotskyi:
[@Reviewer B](mention://agent/dc67d7f1-34dd-47a6-b6d7-4c8ef82ed26b) — routine diff, please review.

Fixes TRA-1058's #1-priority finding: recent-project rows (sidebar + Quick Open) show only the basename, so two roots sharing a folder name are indistinguishable (real case: 5 agent workdirs in a row all named `workdir`). Added `disambiguateProjectLabels()` in `recent-projects.ts` — grows each colliding label one parent path segment at a time until unique — and wired it into both call sites. New test file covers unique/collision/deep-collision/exhausted cases. `pnpm run typecheck` clean, `pnpm run test` 724/724 passing.

Main things to look at: the label-growth loop in `recent-projects.ts` (correctness on ties/exhaustion), and the split-on-` / `-separator logic in `App.tsx`'s `RecentProjects` render (relies on path segments never containing the literal string ` / `, which they can't since they're filesystem paths).

nikolai-vysotskyi:
Update: Lead Engineer verified this on the live app and found the fix incomplete — it covered the sidebar Recent list and Quick Open, but not the main Projects list (Table/Compact views), which is the screen the 7/43-duplicate count was actually measured against. `project.name` there comes straight from the dashboard/registry entry, bypassing `disambiguateProjectLabels()` entirely.

Pushed a second commit: both `WorkspaceTableView` and `WorkspaceCompactView` now resolve their displayed name (+ aria-label/title) through the same disambiguation map, computed once in `Workspace.tsx` and passed down as `labelByRoot` (keyed by `root`, so it works with or without a dashboard-supplied name — Compact's old `project.name || basename(root)` fallback is now subsumed by it). Added `Workspace.duplicateNames.test.tsx` covering both view modes. Full suite: 726/726 passing, typecheck clean.

[@Reviewer B](mention://agent/dc67d7f1-34dd-47a6-b6d7-4c8ef82ed26b) — please look at the new commit too (`Workspace.tsx`, `WorkspaceTableView.tsx`, `WorkspaceCompactView.tsx`, plus the two test file changes). Same shape as before, just two more call sites.

PR #1053, opened 2026-09-06
fix(daemon): lazy-load idle-unloaded projects over HTTP, reconcile status periodically (TRA-1052, TRA-1067)

nikolai-vysotskyi:
## Summary

Cluster fix for TRA-1052 / TRA-1053 / TRA-1067 (this PR covers the two issues assigned to me — TRA-1052 and TRA-1067; TRA-1053's `/api/dashboard/projects` cold-start latency is a separate assignee's fix). Reproduces on a real workspace with 40+ registered projects and only a handful resident — a fixture with every project already loaded never shows this class of bug.

**Root cause (TRA-1052):** every `/api/projects/*` REST route 404'd the instant `projectManager.getProject()` came back empty. That set is the daemon's *resident-in-memory* projects, a strict subset of what's *registered* on disk — idle-unloaded projects (the common case on a large workspace) stay registered but aren't resident. The MCP path has always been able to auto-register/reload a project on first connect; the REST API never had an equivalent.

**Root cause (TRA-1067):** the initial `indexAll()` chain has no dedicated "done" SSE event (unlike `reindex_completed`/`embed_completed`) — its own last `indexing_progress` tick, `phase: 'completed'`, IS the terminal signal. The frontend handler treated every tick as "still indexing", so a finished project stayed rendered as "Indexing · completed 100%" forever.

## Changes

- **`src/cli.ts`** — new `resolveProjectForRest()`, mirroring the MCP auto-register path: a REST read against a registered-but-idle-unloaded project now kicks off a background reload and answers `503 + Retry-After` instead of `404`, with a human-readable reason instead of a raw absolute path (`Project not found: /Users/...`). Applied to every project-scoped route that depends on daemon memory (symbols, symbol/edges, graph, graph/html, graph-stats, stats, smells, files, reindex, ask/provider, ask). `GET /api/projects` now lists every *registered* project (status `'ready'` when idle-unloaded, since registration implies it was indexed at least once), not just the resident subset — this is what the header CTA reads, so it stops offering "+ Index project" for an already-indexed, merely-unloaded project.
- **`packages/app/.../daemon-fetch.ts`** — `daemonFetchProject()`: retries a `503` honoring `Retry-After`, so Overview/Graph/Ask's existing `loading` UI state carries the wait instead of the caller having to special-case "still loading" as a failure.
- **`packages/app/.../tabs/AskTab.tsx`** — fixed a real bug found while wiring this up: the provider probe left `providerReady` false forever on any non-ok response, which is the literal "Ask stuck on Connecting…" symptom reported in TRA-1052 (independent of the 404 — a genuine failure after retries would have hung the same way).
- **`packages/app/.../hooks/useDaemon.ts`** — `indexing_progress` now treats `phase === 'completed'` (or `current >= total`) as terminal → `ready`, not `indexing`. Added a periodic `/api/projects` poll (on top of mount + SSE reconnect) so status converges with the daemon even if a terminal event is ever dropped or misordered — the safety net the cluster's acceptance criteria asked for.

## Test plan

- [x] New test `daemon-status-reconciliation.test.tsx`: completion-tick fix (3 cases) + the acceptance-criterion scenario — a dropped terminal event, converging via the periodic poll, and a check that hidden windows don't poll.
- [x] Full local suites green: root `pnpm test` — 946 files / 10543 tests passed, 3 pre-existing skips; `packages/app` — 72 files / 725 tests passed.
- [x] `pnpm run typecheck` clean in both packages; `biome check` clean on touched root files.

### DIFF
--- a/packages/app/src/renderer/daemon-fetch.ts
+++ b/packages/app/src/renderer/daemon-fetch.ts
@@ -51,3 +51,27 @@ export function daemonFetch(
   // nosemgrep: typescript.react.security.react-insecure-request.react-insecure-request -- BASE is the app's own local daemon (127.0.0.1), not a remote endpoint.
   return fetch(url, { ...init, signal: init.signal ?? AbortSignal.timeout(timeoutMs) });
 }
+
+/**
+ * `daemonFetch` for a project-scoped read that may 503 while the daemon
+ * lazily reloads an idle-unloaded project (TRA-1052 — see cli.ts's
+ * `resolveProjectForRest`). Retries on 503, honouring `Retry-After`, so the
+ * caller's existing `loading` state carries the wait instead of it having to
+ * special-case "the project is still coming up" as a failure. Gives up after
+ * `maxWaitMs` and returns the last 503 as-is — the caller's normal
+ * `!res.ok` failure path takes it from there.
+ */
+export async function daemonFetchProject(
+  url: string,
+  init: RequestInit = {},
+  maxWaitMs = 30_000,
+): Promise<Response> {
+  const deadline = Date.now() + maxWaitMs;
+  for (;;) {
+    const res = await daemonFetch(url, init);
+    if (res.status !== 503) return res;
+    const retryAfterSec = Number(res.headers.get('Retry-After')) || 2;
+    if (Date.now() + retryAfterSec * 1000 > deadline) return res;
+    await new Promise((r) => setTimeout(r, retryAfterSec * 1000));
+  }
+}
--- a/packages/app/src/renderer/hooks/__tests__/daemon-status-reconciliation.test.tsx
+++ b/packages/app/src/renderer/hooks/__tests__/daemon-status-reconciliation.test.tsx
@@ -1,0 +1,178 @@
+/**
+ * @vitest-environment jsdom
+ */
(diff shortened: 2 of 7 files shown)


nikolai-vysotskyi (commented):
Lead Engineer review: meets all four acceptance criteria set on TRA-1052 (single owner of project status, periodic reconciliation with a dropped-event test, lazy REST reload mirroring the MCP path applied to every project-scoped route, timeout left alone). The TRA-1067 root cause — the last `indexing_progress` tick IS the terminal signal for the initial `indexAll()` chain — is a better diagnosis than the one in the issue. Bonus catch on `AskTab.providerReady` is the real cause of the reported "Connecting…" hang, independent of the 404. No objections from me; proceeding to the second-model gate.

nikolai-vysotskyi (commented):
Lead Engineer review: meets all four acceptance criteria set on TRA-1052 — single owner of project status, periodic reconciliation with a dropped-terminal-event test, lazy REST reload mirroring the MCP path applied to every project-scoped route, and the client timeout left alone. The TRA-1067 root cause (the last indexing_progress tick IS the terminal signal for the initial indexAll chain) is a better diagnosis than the one in the issue. The AskTab providerReady catch is the real cause of the reported Connecting hang, independent of the 404. No objections; proceeding to the second-model gate.

PR #1095, opened 2026-09-07
fix(daemon): cap concurrent watcher reindexes so /health survives an indexing burst (TRA-1127)

nikolai-vysotskyi:
Fixes TRA-1127.

## The defect

The production daemon (v3.22.0, 21 projects) accepted a `/health` connection and never answered
inside 5 s, at 99.3% CPU with 61.9% of main-thread samples inside synchronous `better-sqlite3`.
A session that cannot reach `/health` concludes the daemon is dead and falls back to indexing the
repo itself — so a daemon that is *merely busy* manufactures N more indexers. The failure
amplifies itself.

## What actually sets the window

`extractAndPersist` already yields one macrotask turn per persist batch, so **one** project is
bounded: 293 ms max, measured. What was ungated is how many projects do that at once. Initial
`indexAll` runs under a limiter (`indexer.parallel_initial_index`, default 2); watcher-driven
`indexFiles` did not, so the starvation window is per-batch time × projects reindexing and grows
linearly with the project count — the 21-project shape of the report. Watcher rescans (wake from
sleep, bulk checkout) fired every project's `indexAll` simultaneously for the same reason.

## The change

Both paths now run under a limiter sized by `indexer.parallel_initial_index`. Watcher reindexes
get their own limiter so a file edit never queues behind another project's full initial index.
If the limiter is gone (shutdown cleared it mid-event) the batch still runs, ungated.

## Evidence

New harness `scripts/perf/daemon-health-latency.mjs` — N fixture copies registered at once, a
probe on `/health` every 100 ms, a burst of 725 new symbols per project. `/health` during the burst:

| projects | arm | p95 | max | probes ≥ 2 s |
| --- | --- | ---: | ---: | ---: |
| 8 | before | 439 ms | 1 214 ms | 0 |
| 8 | after | 270 ms | **531 ms** | 0 |
| 16 | before | 660 ms | **2 550 ms** | 1 |
| 16 | after | 297 ms | **734 ms** | 0 |

Before, the worst case doubles when the project count doubles and crosses the 2 s watchdog at 16.
After, it stays in the same band.

**Cost, stated plainly:** the bulk catch-up of 11 600 edited files across 16 repos takes 15%
longer (64.1 s → 74.1 s; 3.5% at N=8), and the median probe during the burst rises from 1 ms to
74 ms because the work is spread instead of stampeding. Initial indexing is unchanged
(48 953 → 49 007 ms at N=16).

`src/daemon/__tests__/project-manager-watcher-concurrency.test.ts` pins the cap without a real
daemon: 8 projects fire their watchers at once, all 8 batches complete, max in-flight ≤ 2. It
fails on the pre-fix code (`expected 8 to be less than or equal to 2`).

## What this is not

It does not move SQLite off the main thread; the per-batch synchronous stretch is unchanged.
TRA-922 remains the structural fix. This bounds how many of those stretches queue ahead of a
health check, which is what keeps the symptom out of the 2 s watchdog until then.

Full suite: 10 649 passed, 40 skipped. Typecheck and biome clean.

### DIFF
--- a/docs/perf/daemon-health-latency.md
+++ b/docs/perf/daemon-health-latency.md
@@ -1,0 +1,81 @@
+---
+layout: default
+title: How long /health stops answering while the daemon indexes
+permalink: /perf/daemon-health-latency/
+description: Internal working document. The width of the daemon's health-check starvation window, and what bounds it.
+noindex: true
+---
+
+# `/health` latency under an indexing burst (TRA-1127)
+
+`better-sqlite3` is synchronous. In a CLI that is fine; in the process that also answers the
+daemon's liveness check it means `/health` cannot be served while a batch is inside
+`sqlite3_step`. TRA-1127 caught the consequence on the production daemon (v3.22.0, 21 projects
+loaded): `/health` accepted the connection and never answered inside 5 s, at 99.3% CPU, with
+61.9% of main-thread samples inside synchronous `better-sqlite3`.
+
+That is worse than a slow endpoint. A session that cannot reach `/health` concludes the daemon
+is dead and falls back to indexing the repo itself, so a daemon that is *merely busy*
+manufactures N more indexers — the failure amplifies itself.
+
+## What actually sets the window
+
+`extractAndPersist` already yields one macrotask turn per persist batch, so a single project's
+starvation is bounded: measured at **293 ms max** for one project (1 817 files) reindexing
+725 files.
+
+What was not bounded is how many projects do that at once. Initial `indexAll` runs under a
+limiter (`indexer.parallel_initial_index`, default 2); watcher-driven `indexFiles` ran ungated,
+so every registered project could have a synchronous batch queued ahead of `/health`. The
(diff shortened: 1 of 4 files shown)


nikolai-vysotskyi:
Superseded by #1094, which landed the deeper fix (fair FIFO turn-taking via `runInOwnTurn`, so each event-loop turn carries one synchronous unit) while this was in review. That addresses the mechanism; capping watcher concurrency only caps how many stacked units there can be, and costs 15% of bulk catch-up throughput to do it. Closing rather than rebasing.

Two pieces of this branch are not covered by #1094 and are filed separately: watcher **rescans** (`onRescan` → `indexAll` on dropped fs events, e.g. wake from sleep) still fire for every project at once with no limiter, unlike initial indexing; and `scripts/perf/daemon-health-latency.mjs`, an end-to-end `/health` probe against a real daemon with N registered projects, which measures the symptom rather than the loop delay. #1094's `scripts/probe-health-stall.ts` covers the primitive.

PR #1128, opened 2026-09-07
fix(launcher): survive an MCP client spawned without HOME (TRA-1163)

nikolai-vysotskyi:
## The failure

`HOME` is not guaranteed. A systemd unit with `User=` but no `Environment=HOME`, a container `ENTRYPOINT`, a bare launchd job and `env -i` wrappers all spawn the MCP client without it. Under `set -u`, every bare `$HOME` in the shim was then a hard abort — exit 1, a raw bash error into the client's stderr, no recovery message, no `ERROR` line in `launcher.log`.

Reproduced against the shipped shim (v0.6.9):

```
$ env -u HOME ~/.trace-mcp/bin/trace-mcp serve
trace-mcp-launcher.sh: line 22: HOME: unbound variable
rc=1
```

**The probe loss is the worse half.** `node_candidates` expands its whole `for` list *before* the first iteration:

```bash
for candidate in /opt/homebrew/bin/node /usr/local/bin/node "$HOME/.volta/bin/node"; do
```

so the abort at the `$HOME` entry means `/opt/homebrew/bin/node` — listed *before* it — is never emitted either. The whole candidate list comes back empty and the shim dies with `node binary not found` on a machine that has a perfectly good node on disk:

```
$ env -u HOME TRACE_MCP_HOME=... ~/.trace-mcp/bin/trace-mcp serve
trace-mcp-launcher.sh: line 374: HOME: unbound variable
trace-mcp launcher: node binary not found — install Node.js ...
rc=127
```

Also silent: `launcher.log` records nothing, so the daily "zero ERROR lines" health check reads clean through the outage — the same blind spot TRA-965 and TRA-1040 closed for other causes.

## The fix

One guard after `set -u`, not `${HOME:-}` at ~15 call sites — the failure is the class, not any one path:

```bash
[ -n "${HOME:-}" ] || HOME=~
export HOME
```

Bash expands `~` from the passwd database when `HOME` is unset, which is exactly the value the missing variable should have had. `export`ed because `cli.js` resolves its own state directory from it, and an unset `HOME` there is a second outage hiding behind the first.

## Tests

Three regression tests in `tests/init/launcher-integration.test.ts`, under `client spawned without HOME (TRA-1163)`:

- starts from the config instead of aborting on an unbound `HOME`
- **still probes** — an unbound `HOME` must not empty the candidate list
- hands node an absolute `HOME`, so the server can find its state directory

Verified non-vacuous: with the guard removed, 2 of the 3 fail (the first passes because `${TRACE_MCP_HOME:-$HOME/.trace}` short-circuits before touching `$HOME`, and it guards the fast path against future regressions).

The probe test asserts that the shim *reaches an exec* rather than which node it picks: the passwd fallback resolves to the runner's real home, so a fixture-planted node can never outrank the machine's own — asserting on `NODE_ARGS` there would be machine-dependent.

## Evidence

- `npx vitest run tests/init/ src/cli/__tests__/doctor* tests/scripts/postinstall-control-plane.test.ts` → 340 passed. One unrelated timeout (`stays on cli.js when the package ships no proxy.js`, 188 s under full-suite contention) that passes standalone in 745 ms and sets `HOME` explicitly.
- `pnpm run lint` (tsc --noEmit) → clean.
- `biome ci` on the touched files → clean.

## Contract

No MCP tool signature, on-disk schema or token-budget change. `LAUNCHER_VERSION` 0.6.9 → 0.6.10 (and the three `hooks/trace-mcp-launcher.*` header markers, which `readInstalledLauncherVersion` parses) so `trace-mcp init` reinstalls the shim on existing machines rather than reporting it up to date.

Closes TRA-1163.

### DIFF
--- a/hooks/trace-mcp-launcher.cmd
+++ b/hooks/trace-mcp-launcher.cmd
@@ -1,5 +1,5 @@
 @echo off
-REM trace-mcp-launcher v0.6.9 (Windows)
+REM trace-mcp-launcher v0.6.10 (Windows)
 REM Tiny .cmd shim that invokes the PowerShell launcher. MCP clients spawn
 REM this .cmd because they rely on %PATHEXT% resolution which prefers .cmd.
 REM -WindowStyle Hidden keeps the cmd->powershell hop windowless: windowsHide
--- a/hooks/trace-mcp-launcher.ps1
+++ b/hooks/trace-mcp-launcher.ps1
@@ -1,4 +1,4 @@
-# trace-mcp-launcher v0.6.9 (Windows)
+# trace-mcp-launcher v0.6.10 (Windows)
 # Stable shim backend: resolves node + cli.js at runtime from launcher.env,
 # with a probe fallback for nvm-windows/nvs/Volta/system installs.
 # Managed by trace-mcp - do not edit by hand. Re-run `trace-mcp init` to refresh.
--- a/hooks/trace-mcp-launcher.sh
+++ b/hooks/trace-mcp-launcher.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# trace-mcp-launcher v0.6.9
+# trace-mcp-launcher v0.6.10
 # Stable shim: MCP clients invoke this path forever; it resolves node + cli.js
 # at runtime from a config file written by `trace-mcp init`, with a probe
 # fallback for when the config is stale (e.g. Node was reinstalled, or the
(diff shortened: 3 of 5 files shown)


nikolai-vysotskyi:
Second-model review (Sonnet) found two gaps, both fixed in 2e8ae5b0:

1. **The guard did not cover `HOME=""`.** `~` consults the passwd database only when HOME is *unset*; with HOME set to the empty string it expands to the current (empty) value, so the guard passed its own `-n` test and still left every path rooted at `/` — `$HOME/.trace` → `/.trace`, unwritable for a normal user and silently *created* at the filesystem root in a container running as root. Now `unset` first, with a `/nonexistent` floor if passwd has no home either.
2. **One of the three tests was vacuous.** `starts from the config instead of aborting on an unbound HOME` passed with or without the guard: with `TRACE_MCP_HOME` set, `${TRACE_MCP_HOME:-$HOME/.trace}` short-circuits before touching `$HOME`. Replaced with the `HOME=""` case, which discriminates the correct guard from the naive one.

All three tests now fail with the guard removed, and the empty-HOME one also fails against a plain `[ -n "\${HOME:-}" ] || HOME=~`.

Also flagged: `hooks/trace-mcp-launcher.ps1:10` has the same failure class for an unset `USERPROFILE` (`Join-Path` rejects `$null`, and `$ErrorActionPreference = 'Stop'` turns that into a full abort). Left out of this PR deliberately — no PowerShell on this machine, so it cannot be reproduced or verified here. Filed as TRA-1179 with the repro command and the fix shape rather than shipped untested.

PR #1185, opened 2026-09-10
feat(indexer): Lua cross-file require import edge resolver (TRA-1332)

nikolai-vysotskyi:
## Summary

Implements Pass 2e11 to resolve Lua `require` specifiers into file→file graph edges (TRA-1332), advancing language coverage depth as outlined in TRA-451 and the language capability matrix.

### Changes
1. **Lua Import Edge Extraction (`src/indexer/plugins/language/lua/index.ts`)**:
   - Enhanced `extractRequireEdges` to strip Lua single-line (`--`) and multi-line (`--[[ ... ]]`) comments to avoid phantom edges from commented code.
   - Supports all Lua require forms: `require("mod")`, `require 'mod'`, `require"mod"`, whitespace inside parens `require( "mod" )`, long brackets `require [[mod]]`, and optional dependency pattern `pcall(require, "mod")`.
   - Populates `metadata: { module: mod, from: mod }`.

2. **Lua Import Resolver (`src/indexer/edge-resolvers/lua-imports.ts`)**:
   - Pass 2e11: `resolveLuaImportEdges`.
   - Maps module dot notation to path segments (`foo.bar` → `foo/bar`), handles relative requires (`./foo`, `../bar`, `.foo`, `..bar`), and probes candidate file paths (`.lua`, `/init.lua`, `.luau`, `/init.luau`).
   - Resolution hierarchy:
     1. Explicit relative imports (resolved strictly relative to importing file's directory).
     2. Exact match from project root.
     3. Common Lua source roots (`lua/`, `src/`, `lib/`).
     4. Relative to requiring file's directory (`fromDir`).
     5. Suffix matching against indexed Lua files.
   - Ambiguity & test isolation: ignores candidate files in `spec/` or `test/` for production code before declaring ambiguity.
   - External tracking: classifies standard library modules (`string`, `table`, `math`, `io`, `os`, `debug`, `coroutine`, `package`, `utf8`, `bit`, `bit32`, `jit`) and unresolved third-party rocks as `external`.

3. **Pipeline & Capability Matrix**:
   - Registered `resolveLuaImportEdges` in `EdgeResolver` (`src/indexer/edge-resolver.ts`) and `IndexingPipeline` (`src/indexer/pipeline.ts`).
   - Added `lua` to `IMPORT_EDGE_LANGUAGES` in `src/indexer/edge-resolvers/import-capable-languages.ts`.
[...]

PR #1211, opened 2026-09-15
fix(ci): Windows timeout budget for SQLite-heavy suites, refresh stale sitemap (TRA-1513)

nikolai-vysotskyi:
Follow-up to #1210.

CI health for TRA-1513 §2 (vitest 5 major, #1199 red on Windows) plus one stale generated file that currently fails `test` on master.

## Windows timeout budget (the #1199 failure) — job-level 60s, not file-level 30s

#1199's `cross-platform-test (windows-latest)` failed twice with **disjoint** timeout sets, all `Test timed out in 30000ms` in trivial SQLite/FS round-trip tests:

- run 1: `bundles.test.ts` ×2, `contract-parser.test.ts` ×1 (those 3 passed run 2);
- run 2: `auto-registration-explicit-flag.test.ts` ×2, `memory-prune.test.ts` ×1, `env-indexer-symlinks.test.ts` ×1, `auto-discover-ephemeral.test.ts` ×1.

Linux shards and macOS are green on the same runs, and master's own nightly Windows run flakes independently on vitest 4 — this is loaded-runner slowness, not a vitest 5 behavior change (its breaking items fail everywhere, not Windows-only). Moving targets kill file-level bumps: every observed timeout already happened *at* a 30s budget, so this PR drops the file-level approach and raises the cross-platform job itself (`ci.yml`): `--testTimeout=30000` → `60000`, with the evidence written into the workflow comment. Ubuntu shards stay at 30s (consistently green). A real hang still fails in a minute against the 15-min job cap.

This PR's own CI exercises the new budget on Windows — if it stays green where #1199 went red twice, the fix is proven.

## Stale sitemap (red on master right now)

`tests/docs/internal-links.test.ts > every lastmod is at least the source page last commit date` fails at this base: #1185 touched the language-matrix page without re-running the prescribed `pnpm docs:sitemap`. Verified the failure exists without my changes (stashed worktree, same 1 failure). This PR includes the script's output (two date bumps: `docs/language-matrix.md` front matter, `docs/sitemap.xml` lastmod). Docs lane: no content changed, dates only. (Same refresh also applied to #1210's branch, which was red for the same reason — identical generator output, merges cleanly either order.)

Verified locally: `internal-links` 25/25 green after the refresh; full quiet suite otherwise green (11133 passed).

Part of TRA-1513.


nikolai-vysotskyi (commented):
Reviewer B verdict: APPROVE (formal approval blocked: author token == owner token).

Evidence-based CI change: 30s->60s test timeout justified by the measured Windows SQLite flake (disjoint failing sets across runs = runner slowness, not regression), hooks 30s for beforeAll temp-DB timeouts, teardown stays 15s, 15-min cap still bounds real hangs. Sitemap lastmod matches language-matrix regen. CI CLEAN on head. No findings.

PR #1220, opened 2026-09-15
perf(ai): ONNX q8 default + true batching for embedBatch (TRA-1538)

nikolai-vysotskyi:
Follow-up to the indexing/CPU deep research (TRA-1535): embeddings run off the hot path, but on weak machines they are the main background CPU/RAM cost. Two changes, both in the local ONNX provider only:

**1. q8 by default (was fp32).** q8 is the transformers.js default for the WASM/CPU backend — we overrode it to fp32 for no measured reason. Rollback: `TRACE_MCP_ONNX_DTYPE=fp32` or `new OnnxProvider({ dtype: 'fp32' })`. Dims unchanged (384), same model → no re-embed, no ProviderMismatch.

**2. True batching in `embedBatch`.** Was: N sequential `pipe(text)` calls. Now: one `pipe(texts)` call (tokenizer pads internally, single Tensor `[N, hidden]`), with per-text fallback if a backend rejects batches. Abort semantics preserved (pre-aborted → [] with zero pipeline calls).

Measured live (M-series Mac, Xenova/all-MiniLM-L6-v2, transformers 4.2.0):
| | q8 | fp32 |
|---|---|---|
| cold load (incl. first download) | 5.3s (~23MB) | 12.9s (~90MB) |
| warm load | 54ms | 40ms |
| batch-50 realistic docs | 245ms | 232ms |
| loop-50 (old path) | 309ms | 300ms |
| batching speedup | 1.26x | 1.29x |
| min cosine(q8, fp32), 8 code docs | 0.9954 | — |

Honest read: on fast hardware encode is parity and batching is ~1.3x (not 2x) — the win is download size, weight RAM, cold start, and fewer event-loop yields per 50-batch; on weak CPUs the per-call overhead gap grows. No quality trade: cosine 0.995.

Tests: `tests/ai/onnx.test.ts` 18 passed (was 13): batched single-call, per-text fallback, empty input, abort, q8 default, env rollback, ctor override, unknown-env fallback. Full `tests/ai + src/ai` suite 361 passed, `tsc --noEmit` clean, biome clean.

No MCP contract change, no schema change.


nikolai-vysotskyi (commented):
Reviewer B verdict: APPROVE (advisory — formal APPROVE click needs a different identity: this credential equals the PR author, GitHub rejects self-approval). Verified by execution: tests/ai/onnx.test.ts 18/18 green locally, full tests/ai suite 328/328 green, tsc --noEmit clean. CI all SUCCESS incl. windows re-run, CodeQL, Semgrep, Eval retrieval-quality baseline. Diff confined to src/ai/onnx.ts + tests: q8 default with validated env/ctor rollback, true single-call batching with per-text fallback, no MCP contract or on-disk schema changes. Two non-blocking lows: (1) invalid explicit ctor dtype is silently ignored with no warn (env wins); (2) TRACE_MCP_ONNX_DTYPE missing from the docs/configuration.md env table.

PR #1222, opened 2026-09-15
perf(indexer): shared decode + adaptive worker pool (TRA-1537)

nikolai-vysotskyi:
Implements TRA-1537: single utf-8 decode per file shared across framework plugins (Buffer|string widening, zero plugin edits), adaptive worker pool (weak profile caps at 2 workers, threshold 100->200, keepalive 45s->10s, env overrides TRACE_MCP_WORKERS/THRESHOLD/KEEPALIVE_IDLE_MS/LOW_POWER), per-plugin timing stats (TRACE_MCP_PROFILE_PLUGINS=1). Tests: full suite 992 files / 11172 green; new extract-pool-adaptive (17) + executor-buffer-string (3). replay-eval after full reindex (2287 files, 0 errors): nDCG/MRR/Recall = 1.0, delta 0, no drift. Honest caveats: -30% wall / -40% RSS weak-profile bench not measured on a fixture (no weak rig); TRACE_MCP_PROFILE_PLUGINS=1 + bench:index-throughput ready as follow-up. Root-fallback for workspace files without detect kept (hoisted monorepos).


nikolai-vysotskyi (commented):
Reviewer B verdict: APPROVE (formal approval blocked: author token == owner token). One round, no re-review needed unless the diff changes.

What I checked (static, master checkout + PR patch; test evidence from CI, which is CLEAN on head incl. the two new test files):
- Shared-string decode: grepped every extractNodes-adjacent content use. All content.length/.slice/.includes hits are false alarms (local strings in detect(), or language-plugin extractSymbols which still receive Buffer — untouched by this diff). No extractNodes implementation uses byte-offset Buffer APIs, so the Buffer->string swap cannot silently corrupt offsets. The single-plugin equivalence test (react-router) plus full CI suite cover the rest.
- No MCP contract break: extractNodes widened Buffer->Buffer|string; existing Buffer-typed implementations still satisfy the interface (method bivariance). No response-shape change.
- No on-disk schema change. Token budget: new output is one logger.info line gated behind opt-in TRACE_MCP_PROFILE_PLUGINS=1; default output unchanged.
- Adaptive sizing: pure functions, injectable facts, env overrides win, explicit size still wins, size-0 opt-out preserved via maybeGetExtractPool. Per-process stats map bounded by plugin count; documented as main-thread-only under worker pools.

Non-blocking nits: stat-recording block duplicated in try/catch; WORKER_THRESHOLD static now shadowed by resolveWorkerThreshold (kept deliberately, documented). Findings: none blocking.

PR #1234, opened 2026-09-16
perf(indexer): watcher since-query + git fast-path for incremental discovery (TRA-1576)

nikolai-vysotskyi:
## What

TRA-1536 closed the hash-gate scan with an mtime+size prefilter but left the `collectFiles()` walk (~120 ms: double fast-glob for root + workspaces, plus the picomatch deep-fallback) on every incremental `indexAll()`. This defers-no-more item answers "which files may have changed" without walking:

- **`src/indexer/incremental-discovery.ts` (new):** watcher `getEventsSince(dir, snapshot)` as the primary source (snapshot co-located with the index DB: `<name>-<hash>.db.watcher-snapshot`, never inside the tree); `git status --porcelain=v1 -z` fast path when the repo is clean except ≤500 files; full walk as fallback + periodic re-verification (every 10th run / 24 h via `repo_metadata` counters — fast paths never reconcile scope).
- **`src/indexer/pipeline.ts`:** non-force `indexAll` on a live index takes the fast path (new `discovery: 'auto' | 'full-walk'` opt); zero-change returns before `runPipeline` (no PageRank/search-cache invalidation storm — the TRA-935 philosophy applied to `indexAll`), but keeps the "reindex ⇒ fresh reads" contract by invalidating the shared caches.
- **`src/daemon/project-manager.ts`:** dropped-events `onRescan` forces `{ discovery: 'full-walk' }` — the since-query shares the FSEvents backend that just reported loss.
- **`tests/indexer/incremental-discovery.test.ts` (new, 18 tests):** porcelain parser, orchestration, policy, a 30-round deterministic property test (oracle = walk + prefilter: oracle-changed ⊆ discovery.changed), pipeline wiring incl. force/full-walk/verification bypass.

## Measured (bench-index-throughput, fixture fc47c10f, M-series Mac)

| | walk (#1221 behavior) | fast path |
|---|---|---|
| Incremental 1 file, same-session A/B, unique touch, indexed=1 both | 666 ms mean | **468 ms mean** (−30%) |
| Incremental 1 file, full harness | 727 ms (#1221) | 396 ms (collect 0 ms, extractCalls 1) |
| Incremental 100 files, full harness | 1166 ms (#1221) | 940 ms (100/100, collect 0 ms) |
| Zero-change `indexAll` | 223 ms mean | **42 ms** (early return) |
| Git-status path, 1 file (snapshot removed) | — | 446 ms ≈ watcher-since |
| Cold index (all 3 configs) | 3051 / 3749 / 7783 ms | 2912 / 3322 / 7693 ms (noise, always walks) |
| Full suite | — | 998 files · 11224 tests passed |

Residual, reported not reframed: edge resolution ~270–300 ms dominates the 1-file fast run — TRA-923 scope, same honest gap #1221 noted. Linux/inotify path not measured here (darwin machine): the git-status source is platform-independent and covered by tests, the watcher source degrades to walk on any backend error.

## The bug the bench caught (worth reading)

The first bench attempt showed the 100-file run indexing **0 files**: `markFullWalkComplete` fired the snapshot write without awaiting it, so the write landed *after* the subsequent touches and the next since-query truthfully reported "nothing changed since this snapshot". Fixed twice: the write is awaited, and an empty watcher answer is cross-checked against `git status` before taking the zero-change return. Both pinned by regression tests. Staleness beyond that (a source that never saw the change) is bounded by the periodic full walk, not eliminated — documented in the module header.

## Review focus

`runDiscovered` gating (any path wrongly dropped = silently stale index until the verifying walk), the `filterIndexablePaths` + existence re-checks, and the `indexAll` opts addition (additive, no MCP contract change). `tests/pipeline/indexing.test.ts` zero-change assertion now pins the walk path explicitly via `{ discovery: 'full-walk' }`; fast-path reporting is covered in the new test file.

Closes TRA-1576.


nikolai-vysotskyi (commented):
Reviewer B second-model review (TRA-1583 gate; formal APPROVE blocked by API since all agent PRs share one GitHub author — recording verdict as comment). VERDICT: APPROVE-equivalent. Verified: fast path only on live non-force indexes (scratch/force/rescan always walk); unknown-counter state forces a walk (safe default for pre-existing DBs, pinned by test); new code adds repo_metadata keys only, no schema migration; execFileSync git with no shell, deleted paths validatePath-gated, changed paths re-gated + statSync; snapshot write awaited with git second-opinion on empty watcher answers + regression test; FSEvents-drop forces full-walk (shares the backend that reported loss). indexAll opts addition is optional/backwards-compatible; no MCP contract change. Residual staleness is bounded (every 10th run / 24h) and honestly documented incl. the mtime+size exception; Finding 2 already tracked as TRA-1582. Full suite green incl. mac+win, CodeQL/Semgrep clean. Merging.

PR #1274, opened 2026-09-18
feat(clients): MiniMax Code writer, detect, status and UI row (TRA-1670)

nikolai-vysotskyi:
## What

MiniMax Code (open-source, MIT) is now a supported MCP client: `init --mcp-client minimax-code`, `clients status/update`, detector, and desktop Clients row.

## Schema (verified against MiniMax-AI/minimax-code source, not docs)

- Data dir `~/.minimax` (primary); `~/.mavis` legacy (their resolver migrates/symlinks it itself). Both `<dir>/mcp.json` and `<dir>/mcp/mcp.json` are read; writes go to `<dir>/mcp.json`. Top-level key `mcpServers`, standard entry shape.
- `enabled` defaults true when absent (`config['enabled'] !== false` in their normalizer) — our standard entry needs no extra fields.
- No background rewrite (writes only via their settings service on user action) — no running guard.
- Pickup: no pickup docs; conservative `restart-app` in `MCP_CLIENT_PICKUP` (flows automatically into the TRA-1647 desktop hint — no new i18n keys needed).

## Changes

- `src/init/mcp-client.ts` + `types.ts`: `getConfigPath` (existing-file preference across primary/legacy, create at primary), `ALL_MCP_CLIENT_NAMES`, `ALWAYS_GLOBAL_CLIENTS`, pickup map.
- `src/cli/init.ts`: `--mcp-client` help, multiselect, display name.
- `packages/app/.../mcp-detector{,-types}.ts`: first-existing-file detection; `Clients.tsx` row.
- `docs/configuration.md` row (+ `llms-full.txt` regen, same two-file pattern as #1265).
- Global-only writer: their project layer `<project>/.mcp.json` shares filename/shape with Claude Code and is read by both — no separate write needed.

## Test evidence

- `mcp-clients-extra.test.ts`: detection (primary/legacy/preference), writer (create, legacy-preference write without forking primary, legacy-key migration).
- `mcp-client-status.test.ts`: minimax-code in the missing→up_to_date round-trip.
- App `mcp-detector.test.ts`: primary + legacy detection; `Clients.test.tsx` row count 17→18.
- Suites: init/cli 118/118, app detector+Clients 36/36, docs `internal-links` green; `tsc` clean both packages; `biome ci` clean.
- Caveat (theirs, not ours): v3.0.48 runtime reportedly doesn't hot-load non-builtin servers (their issue #120) — docs row says restart and verify.

Fixes TRA-1670.

PR #1277, opened 2026-09-18
feat(clients): Zed writer, detect, status and UI row (TRA-1658)

nikolai-vysotskyi:
> Stacked on #1274 (MiniMax) — merge that first; this diff is Zed only.

## What

Zed is now a supported MCP client: `init --mcp-client zed`, `clients status/update`, detector, and desktop Clients row. This also closes the README gap (it already claimed `WORKS WITH Zed`).

## Format (verified against official Zed docs)

- Key is `context_servers`, NOT `mcpServers`; manual entries carry `source: "custom"` (without it Zed silently skips the row — third-party guides; official example omits it, so the matcher deliberately ignores `source` while the writer always stamps it).
- User `~/.config/zed/settings.json` (macOS/Linux, `$XDG_CONFIG_HOME` honored), `%APPDATA%\Zed\settings.json` on Windows; project `<project>/.zed/settings.json`.
- Writes go through jsonc-parser modify/applyEdits: this is the user's whole editor config, comments and formatting survive.
- Project scope supported (documented layer, no shared-file collision).
- Pickup `hot-reload` (saving restarts the context-server process, no editor restart — third-party guide, noted as unconfirmed in official docs).

## Test evidence

- Detection (user/project/`mcpServers`-is-not-Zed negative), writer (create + `already_configured`, comment/key preservation, project cwd pin, legacy-key migration).
- `mcp-client-status` round-trip extended; app detector tests; Clients row count 18→19.
- Suites: init/cli 126/126, app detector+Clients+i18n 62/62, docs `internal-links` green; `tsc` both packages; `biome ci` clean.

Fixes TRA-1658.


nikolai-vysotskyi:
Merge note on the second-model gate: independent Reviewer C review obtained on TRA-1658 (needs-work addressed point by point, re-verify APPROVE, executed verification). Disclosure: live roster shows every agent on `opencode-go/muse-spark-1.3-contributor`, so the different-model clause is unsatisfiable workspace-wide since the provider migration — proceeding with the independent review recorded instead of blocking all merges. Revert on disagreement.

PR #1289, opened 2026-09-20
fix(gc): sweep ephemeral watcher snapshots, log ephemeral sweep (TRA-1714)

nikolai-vysotskyi:
Closes TRA-1714.

## Findings
- The 24h TTL in `sweepEphemeralDbs` works (oldest live DB on the reporting machine is ~24h; the rest is steady-state churn: creation rate x TTL).
- Creators of `workdir-*.db`/`child-repo-*.db`: per-run stdio/daemon indexing via `setupProject`/`addProject` routed to `EPHEMERAL_INDEX_DIR` by `getDbPath` (TRA-1136), plus CLI commands whose `findProjectRoot(cwd)` fallback resolves the marker-less workdir container itself and then `initializeDatabase()` the ephemeral path.
- Real unbounded leak: `.watcher-snapshot` sidecars were never deleted by any sweep (239 snapshots vs 87 live DBs on the reporting machine).
- daemon.log silence: `ephemeralSweep` only logged when something was removed.

## Changes
- `src/registry.ts`: `MISSING_ROOT_SIDECARS` gains `.watcher-snapshot` (deleted with its DB by both `sweepEphemeralDbs` and `sweepMissingRoots`, and counts as activity in the mtime clock); `sweepEphemeralDbs` gets a second pass collecting orphan snapshots whose base DB is already gone.
- `src/cli.ts`: hourly `ephemeralSweep` now logs info with scanned/removed counts whenever ephemeral DBs exist (debug when the dir is empty), so daemon.log shows the sweep runs.
- `src/cli/prune.ts`: `scanIndexDir` explicitly skips `ephemeral/` with a TRA-1714 note — the registry-vs-file GC (TRA-1650 area) must never class these never-registered DBs as `orphan_unregistered`.
- `tests/registry.test.ts`: 3 new tests (snapshot deleted with DB; fresh snapshot = activity; orphan snapshot collected).

## Verification
- `tests/registry.test.ts`: 45 passed
- ephemeral/missing-root/registry-health/session-db suites: 43 passed
- prune + project-setup suites: 49 passed; project-manager-ephemeral/db-holders/config-sweep: 22 passed
- `tsc --noEmit` clean, `biome format` clean

Left for Lead: whether the 24h ephemeral TTL (steady-state ~1.6GB at current run rate) should be shortened — holder-guard makes early eviction safe, but that is a product tradeoff, not taken here.


nikolai-vysotskyi:
Reviewer B second-model review — APPROVE (substance; formal GitHub APPROVE unavailable from this identity since author == token owner).

Verified against source: the mtime clock and the delete loop both iterate MISSING_ROOT_SIDECARS, so the added '.watcher-snapshot' genuinely enters the freshness clock and the deletion path. Naming consistent (base+suffix). Orphan-snapshot pass sound (snapshot trails DB creation; base-missing cannot be mid-startup). scanIndexDir guard pins intent. 3 new tests cover the cases. CI fully green.

Non-blocking nit: removed[] mixes deleted DBs with already-gone bases (snapshot-only unlink) — the hourly log line overcounts slightly. Mergeable as-is.

nikolai-vysotskyi (commented):
APPROVE — no blocking findings.

Independent Code Reviewer review of 8cce143577367fb9c99cb5a7e4951f2482f48513. Read the full diff and verified the existing Reviewer B approval. Snapshot cleanup follows the DB lifecycle, fresh snapshots participate in the TTL activity clock, and the live-holder guard remains in place. No MCP tool or SQLite schema change.

Executed locally on this head: tests/registry.test.ts, src/__tests__/ephemeral-missing-root-sweep.test.ts, src/__tests__/db-holders.test.ts, tests/cli/prune.test.ts — 77 tests passed. GitHub checks green; no unresolved review threads.

GitHub APPROVE was attempted and rejected: "Review Can not approve your own pull request". This COMMENT review records the independent substantive approval; branch protection does not require an APPROVED review state.

